PR #4, opened 2023-11-01
fix: Remove the empty line at the top of generated files

arnevm123:
Hi,
I noticed the generated files had empty lines at the top of them.
I fixed this in my PR. It does make the actual code a bit less pretty, but I think that's a fair trade-off


Melkeydev (approved):
lgtm!

PR #7, opened 2023-11-01
feat: Add gofmt after creating the go files

arnevm123:
Not sure if this is something you want in this project, but this would make the whitespace problems a non-issue, but would make the generation of the files a bit slower.

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -172,6 +172,12 @@ func (p *Project) CreateMainFile() error {
 		return err
 	}
 
+	err = utils.GoFmt(projectPath)
+	if err != nil {
+		log.Printf("Could not gofmt in new project %v\n", err)
+		cobra.CheckErr(err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


cmd/program/program.go
```diff
@@ -172,6 +172,12 @@ func (p *Project) CreateMainFile() error {
 		return err
 	}
 
+	err = utils.GoFmt(projectPath)
+	if err != nil {
+		log.Printf("Could not gofmt in new project %v\n", err)
+		cobra.CheckErr(err)
+	}
+
```
Melkeydev, line 180: Correct me if i am wrong but wouldnt this always fail for users who do not have gofmt installed on their computer/env?

Melkeydev (commented):
Adding a comment on gofmt

cmd/program/program.go
arnevm123: Yes and no.
It would fail if gofmt is not installed, but when installing go for the first time gofmt should be installed as well.
If i look in a tar file here: https://go.dev/dl/ I can see that in the go/bin file there's a go and gofmt binary present.

Melkeydev: Good point. 
I think I am OK to add this then!

Melkeydev (approved):
LGTM

PR #8, opened 2023-11-01
feat: add echo support

### LINKED ISSUE #5: Add echo support
It would be nice to have the option to start a project with [echo](https://echo.labstack.com/docs/quick-start) as well.  I could give it a try if you don't mind.

xsadia:
closes #5 

adds support to the [echo web framework](https://echo.labstack.com/)


cmd/template/echoServer.go
```diff
@@ -0,0 +1,70 @@
...
+		"{{.ProjectName}}/internal/server"
+	)
+
+	func main() {
+		server := server.New()
+		server.RegisterEchoRoutes()
+		server.Logger.Fatal(server.Start(":8080"))
+	}
+`)
+}
```
Melkeydev, line 70: I dont think Echo needs a new server instance - just the routes.
Echo follows a similar pattern to Gin, where it uses a new instance and requires its own context in the parameters.
Look here for how Gin was implemented:
https://github.com/Melkeydev/go-blueprint/blob/main/cmd/template/ginRoutes.go#L15


I think we can follow the same and just create a MakeEcho Routes and re-use the same HTTP server

Melkeydev (changes requested):
Add review for implementation

Melkeydev:
While we are at it, maybe we can also include the echo middleware as well?
```
 // Middleware
  e.Use(middleware.Logger())
  ```

File: cmd/utils/utils.go [hunk 2]
```diff
@@ -26,11 +26,13 @@ func InitGoMod(projectName string, appDir string) error {
...
-		[]string{"get", "-u", packageName},
-		appDir); err != nil {
-		return err
+func GoGetPackage(appDir string, packages []string) error {
+	for _, packageName := range packages {
+		if err := ExecuteCmd("go",
+			[]string{"get", "-u", packageName},
+			appDir); err != nil {
+			return err
+		}
```
xsadia, line 35: had to change this function to support the echo middleware lib. I assumed we didn't have a way to get more than one package per framework @Melkeydev

Melkeydev: @xsadia You actually dont need to explicitly do a go get for middleware.
```
go get github.com/labstack/echo/v4
```

and then just having the imports as:
```
import (
  "github.com/labstack/echo/v4"
  "github.com/labstack/echo/v4/middleware"
  "net/http"
)
```
Should work

xsadia: When I create a project without explicitly doing a go get for the middleware I get 

```
error while importing github.com/labstack/echo/v4/middleware: missing go.sum entry for module providing package github.com/golang-jwt/jwt (imported by github.com/labstack/echo/v4/middleware); to add:
go get github.com/labstack/echo/v4/middleware@v4.11.2
```
I don't know if I missed something @Melkeydev

Melkeydev: :thinking: Oh interesting - the Echo read me is a little misleading then. 
NVM my previous comment. I am going to pull this PR down and check it out later today (looking good though)

xsadia: Ok! Just ping me if you need anything

cmd/template/echoRoutes.go
```diff
@@ -0,0 +1,43 @@
...
+	import (
+		"net/http"
+	
+		"github.com/labstack/echo/v4"
+		"github.com/labstack/echo/v4/middleware"
+	)
+func (s *Server) RegisterRoutes() http.Handler {
+	e := echo.New()
+	e.Use(middleware.Logger())
+	e.Use(middleware.Recover())
```
itschip, line 28: You can just call `echo.Default()` as it uses Logger and Recovery by default :P

Melkeydev: @xsadia Ah this is a good point
If we use echo. Defauflt() it comes with the logger from middleware 
so we potentially dont even need to include it as a second go get dep

itschip: I may or not have been malding while looking at this and confused myself with Gin..

Melkeydev (approved):
LGTM!
ECHO ADDED

Issue #16, opened 2023-11-02
Add full on CLI support


arnevm123:
Instead of opening the TUI, could we add full on CLI support?
In just one command just make a new project:

`go-blueprint --name name --framework Chi`

Melkeydev:
This is a really good option to have for people.
i dont want to replace the current CLI but this gives users an excellent option for faster spin-up

PR #18, opened 2023-11-02
add flag support

itschip:
Adds initial flag support for name and type (#16 )


itschip:
Seems like gofmt went slightly wild here

cmd/create.go
```diff
@@ -29,12 +29,16 @@ const logo = `
...
+	logoStyle           = lipgloss.NewStyle().Foreground(lipgloss.Color("#01FAC6")).Bold(true)
+	endingMsgStyle      = lipgloss.NewStyle().PaddingLeft(1).Foreground(lipgloss.Color("170")).Bold(true)
+	allowedProjectTypes = []string{"chi", "gin", "fiber", "gorilla/mux", "httprouter", "standard-library", "echo"}
 )
 
 func init() {
 	rootCmd.AddCommand(createCmd)
+
+	createCmd.Flags().StringP("project-name", "n", "", "Name of project to create")
+	createCmd.Flags().StringP("project-type", "t", "", fmt.Sprintf("Type of project to create. Allowed values: %s", strings.Join(allowedProjectTypes, ", ")))
```
Melkeydev, line 41: I have one concern with this - if we want to have flag support we should not allow the CLI to kick off.
It should be one or the other, otherwise we can introduce some nasty BUGS

Melkeydev:
Can you add an example of how to run this BTW? Like what are the flags that should be passed in, and potentially add that to the documentation?

cmd/create.go
itschip: The reasoning behind not splitting up as if you only place the name, we could still show the TUI for selecting the type of project to configure. We could still build it in another way tho, for sure!

itschip:
> Can you add an example of how to run this BTW? Like what are the flags that should be passed in, and potentially add that to the documentation?

ex. `go-blueprint create -n testprojet` will set the name and show the UI for select project

`go-blueprint create -n testproject -t gin` create a gin project called `testproject`

cmd/create.go
Melkeydev: Hold on I just tested this out and its super nice.

Melkeydev, line 41: Instead of -t can we make it -f

Melkeydev:
Giving just a project name and no framework causes a bug :cry:

README.md
```diff
@@ -9,7 +9,9 @@
 
 ### 🚀 Introducing the Ultimate Golang Blueprint Library 🚀
 
-Are you tired of starting Golang projects from scratch, struggling with inconsistent structures and code conventions? Go Blueprint is a CLI tool that allows users to spin up a Go project with the corresponding structure seamlessly. It also gives the option to integrate with one of the more popular Go frameworks!
+Are you tired of starting Golang projects from scratch, struggling with inconsistent structures and code conventions? Go
+Blueprint is a CLI tool that allows users to spin up a Go project with the corresponding structure seamlessly. It also
```
benjiewheeler, line 13: I believe you have a few extra newlines here

Melkeydev (approved):
LGTM

PR #19, opened 2023-11-02
FEATURE: Generate README.md on project creation

### LINKED ISSUE #14: Generate a README in the new project with list of commands
We should include a read me that explains the commands from the Makefile

nhlmg93:
Resolves #14


cmd/template/main.go
```diff
@@ -48,3 +48,33 @@ clean:
...
 		`)
 }
+
+func ReadmeTemplate() []byte {
+	return []byte(
+		`
+# Project Title
+
+One Paragraph of project description goes here
+
```
Melkeydev, line 58: An idea here would be to template the ProjectName from the struct into this so instead of Project Title we can use the project name here
```{{.Pojrectname}}```

File: cmd/template/main.go [hunk 2]
```diff
@@ -48,3 +48,33 @@ clean:
...
+make build
+
+run the application
+make run
+
+run the test suite
+make test
+
+clean up binary from the last build
+make clean
```
Melkeydev, line 78: Should we consider putting these commands in a code block?

Melkeydev (changes requested):
Left comments

cmd/template/main.go
nhlmg93: i tried to figure that one out.... you see the issue is the ` character. the template literal string thinks its being delaminated due to markdown using 

```
code block
```

new to go so was not sure how to fix.

cmd/template/main.go
nhlmg93: see latest commit

cmd/template/main.go
arnevm123: This is a possible workaround:

````
`
## MakeFile

run all make commands with clean tests
` + "```bash" + `
make all build
` + "```" + `

build the application

` + "```bash" + `
make build
` + "```" + `
`
````

nhlmg93: see latest commits

Melkeydev (approved):
LGTM

PR #21, opened 2023-11-03
Feature: Initialize Git repository

20pa5a1210:
Added a command to execute `npx gitignore go`. The command generates a .gitignore file configured for Go projects.


20pa5a1210:
# Generated .gitignore file
image

cmd/program/program.go
```diff
@@ -172,7 +171,21 @@ func (p *Project) CreateMainFile() error {
...
+        return err
+    }
+    // Create gitignore
+    err = utils.ExecuteCmd("npx", []string{"gitignore", "go"}, projectPath)
+    if err != nil {
+        log.Printf("Error creating gitignore: %v", err)
+        cobra.CheckErr(err)
+        return err
+    }
+    return nil
```
Melkeydev, line 188: What if the user does not have NPX installed? 
I am wondering if it makes more sense to follow the exciting template pattern with a .gitignore file that we can create and then inject?

20pa5a1210: yeah npx could be a problem. i will use existing template pattern to inject .gitignore file

20pa5a1210:
I made changes by injecting a .gitignore file instead of using NPX

cmd/program/program.go
benjiewheeler: the npx gitignore command simply downloaded the chosen gitignore file from this repo https://github.com/github/gitignore

This can be done in pure Go without NPX

cmd/template/main.go
```diff
@@ -48,6 +48,28 @@ clean:
...
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+# Dependency directories (remove the comment below to include it)
+# vendor/
+
+# Go workspace file
+go.work
+		`)
```
Melkeydev, line 72: You are missing a closing } bracket on this function that is causing 
```go build``` to error and fail out

**Review by Melkeydev — CHANGES_REQUESTED** [review]
Go build failing out

**Review by Melkeydev — APPROVED** [review]
LGTM
I am approving but I will give you the option to merge because the current default branch is master.
I will leave it to you if you want to change it to main. If you do re-request my review, otherwise merge ahead!

File: cmd/program/program.go [hunk 3]
```diff
@@ -192,6 +191,38 @@ func (p *Project) CreateMainFile() error {
...
+		cobra.CheckErr(err)
+		return err
+	}
+    // Switch to main branch
+    err = utils.ExecuteCmd("git", []string{"branch", "-M", "main"}, projectPath)
+    if err != nil {
+        log.Printf("Error switching to main branch: %v", err)
+        cobra.CheckErr(err)
+        return err
+    }
```
20pa5a1210, line 208: checkouts to main branch after git init

20pa5a1210:
i made changes that now git branch defaults to main not master

Melkeydev:
Unfortunately:
```2023/11/07 12:11:07 Error switching to main branch: exit status 128
Error: exit status 128
```
This is what happens when I spin up a new project

Melkeydev (changes requested):
Spinning up new project is failing - please see comment

20pa5a1210:
image
 
## this error happens because when user has default settings to initialize a repo as main
image

## so we force the branch to rename as main then it will not error out

20pa5a1210:
image
## if we add `-f` flag its force rename it and even if it is main by default

20pa5a1210:
## tested and its working fine for me (once test on your end also)
[goprinttest.webm](https://github.com/Melkeydev/go-blueprint/assets/95855318/84a3437f-ca01-4f1b-be75-8782bd7a01de)

Melkeydev (changes requested):
I am not sure what i am doing wrong but:
I pulled your fork, fetched the latest commits and tried to do the same with the same error:
```
2023/11/09 13:22:20 Error renaming master branch to main: exit status 128
Error: exit status 128
```

20pa5a1210:
so i cant figure out the issue with switching to main branch. it working on my side (i tested on my friends machine it worked fine) 
anyway it not working on your side (i assume it has to do with some git issue or  setting). so for now its good to stick with master 
branch .so  i removed `git branch -m main ` cmd.

basokant:
> i removed `switching to main branch cmd`. you can review changes and merge if you think sticking with master branch is ok

👍 I think this is fine too, especially considering that running git init will use the default branch setting on the user's machine, which most people have set to main anyway. And it's easy enough for the user to rename the branch anyway. Good work on the PR!

Melkeydev (approved):
LGTM!

Issue #11, opened 2023-11-02
Add opt-in DB connection structure 


xsadia:
Setting up a database and connecting to it in Go projects is a trivial task most of the time, but, similar to creating an HTTP server, it becomes repetitive. My proposal is to introduce an opt-in database setup.

Melkeydev:
This is a super nice feature we should include

MohammadAlhallaq:
Are any of you guys working on this, or should I give it a try?

Melkeydev:
> Are any of you guys working on this, or should I give it a try?

It's all you bro!

itschip:
Suggestions for modules:
[GORM](https://gorm.io/)
[Bun](https://bun.uptrace.dev/)

And ofc the std sql, and sqlite

silaselisha:
what about providing an option to pick the type of DB either NoSQL or a relational database as mentioned, std SQL.

PR #22, opened 2023-11-03
adding DB support

MohammadAlhallaq:
#11


MohammadAlhallaq:
Hey @Melkeydev,

would appreciate your feedback on the implementation, let me know what could have been done better or if there's anything I might have missed in it.

.gitignore
Melkeydev, line 1: We should probably remove this file from the PR

README.md
```diff
@@ -41,8 +41,15 @@ go-blueprint create
...
-- [Gorilla/mux](https://github.com/gorilla/mux)
+- [HttpRouter](https://github.com/mattn/go-sqlite3)
+- [Gorilla/mux](https://github.com/mongodb/mongo-go-driver)
+
+### Database Drivers Supported
+
+- [Mysql](https://github.com/go-sql-driver/mysql)
+- [Postgres](https://github.com/lib/pq)
+- [Sqlite](https://github.com/gofiber/fiber)
+- [Mongo](go.mongodb.org/mongo-driver)
```
Melkeydev, line 52: jheeze

cmd/template/DBDriver/mongo.go
```diff
@@ -0,0 +1,28 @@
+package DBDriver
+
+type MongoTemplate struct{}
+
+func (m MongoTemplate) Service() []byte {
+	return []byte(`
+import (
```
kevburnsjr, line 7: Missing package name

cmd/template/DBDriver/mysql.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (m MysqlTemplate) Service() []byte {
+	return []byte(`package services
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	_ "github.com/go-sql-driver/mysql"
+	"log"
+	"time"
```
kevburnsjr, line 13: Recommend grouping imports
```go
import (
	"stdlib"

	"external"

	"internal"
)
```

ie.

```go
import (
	"context"
	"database/sql"
	"fmt"
	"log"
	"time"

	_ "github.com/go-sql-driver/mysql"
)
```

cmd/template/DBDriver/mysql.go
```diff
@@ -0,0 +1,47 @@
+package DBDriver
+
+type MysqlTemplate struct{}
+
+func (m MysqlTemplate) Service() []byte {
+	return []byte(`package services
+import (
```
kevburnsjr, line 7: `go fmt` - Missing newline between package name and import statement

cmd/template/DBDriver/mysql.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (s *Service) Health() {
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+
+	err := s.db.PingContext(ctx)
+	if err != nil {
+		fmt.Errorf(fmt.Sprintf("db down: %v", err))
+		return
```
kevburnsjr, line 43: unnecessary return

cmd/template/DBDriver/postgres.go
```diff
@@ -0,0 +1,41 @@
...
+	_ "github.com/lib/pq"
+	"log"
+	"time"
+)
+
+type Service struct {
+	db *sql.DB
+}
+
+func New() *Service {
```
kevburnsjr, line 20: Recommend using interfaces for mockable services
```go
type Service interface {
	Health()
}

type service struct {
	db *sql.DB
}

func New() *service {
	// ...
}
```

newtoallofthis123:
I'm planning on improving upon this by adding some prompts for a .env file? I mean that is I mostly use DB drivers....so would that be cool?

newtoallofthis123:
I could also probably add some fields like

```
DB_URL=postgres://
DB_NAME=name
```

I could use the popular https://github.com/joho/godotenv package?

Melkeydev:
> I could also probably add some fields like
> 
> ```
> DB_URL=postgres://
> DB_NAME=name
> ```
> 
> I could use the popular https://github.com/joho/godotenv package?

Do we need to introduce another package to lead .env vars or can we just use the standard os lib?
Either way, would it make sense to make it into a separate PR?
I am planning on reviewing just this PR and how the DB integration is set up tomorrow

newtoallofthis123:
True....we can probably just read of the os lib as a default, but have an option of a .env? moreover, godotenv is quite a popular package with good reputation

cmd/template/DBDriver/postgres.go
MohammadAlhallaq: Wouldn't the IDE start yelling that we are trying to return an unexported type here?

Melkeydev:
> True....we can probably just read of the os lib as a default, but have an option of a .env? moreover, godotenv is quite a popular package with good reputation

Good points I am OK to add godotenv to either this repo or a following one

Melkeydev:
@MohammadAlhallaq There are some new merge conflicts due to previous merges!

cmd/program/program.go
```diff
@@ -128,6 +166,31 @@ func (p *Project) CreateMainFile() error {
 		}
 	}
 
+	if p.DBDriver != "None" {
```
Melkeydev, line 169: We should probably also integrate the NEW CLI flag feature here as well from this PR:
https://github.com/Melkeydev/go-blueprint/pull/18

i will let you decide if you want to integrate this into this PR or a separate one!

cmd/program/program.go
```diff
@@ -216,6 +279,10 @@ func (p *Project) CreateFileWithInjection(pathToCreate string, projectPath strin
 	case "routes":
 		createdTemplate := template.Must(template.New(fileName).Parse(string(p.FrameworkMap[p.ProjectType].templater.Routes())))
 		err = createdTemplate.Execute(createdFile, p)
+
+	case "services":
+		createdTemplate := template.Must(template.New(fileName).Parse(string(p.DBDriverMap[p.DBDriver].templater.Service())))
```
Melkeydev, line 284: The current set up creates this in a 
```
root/services
```
folder.

I am wondering to keep project structure, can we add it to the existing internal directory and create a new sub-folder called database?
```
/internal/database/database.go
```

Melkeydev (changes requested):
Added a few comments - tested out the PR everything seems to work however,
the DB doesnt actually get hooked up.
Wondering if there is a way we can fully test out this feature and implement it for a user to use

cmd/steps/steps.go
```diff
@@ -54,6 +55,18 @@ func InitSteps(options *Options) *Steps {
...
 			},
+			{
+				StepName: "Go Project Database Driver",
+				Options: []Item{
+					{Title: "Mysql"},
+					{Title: "Postgres"},
+					{Title: "Sqlite"},
+					{Title: "Mongo"},
+					{Title: "None"},
+				},
```
Melkeydev, line 66: Nit: do you think it is worth adding a description to these?

cmd/template/DBDriver/postgres.go
kevburnsjr: Nope. The calling code would depend on the interface rather than the concrete struct type.

```go
type healthHandler struct {
	s services.Service
}

func NewHealthHandler() *healthHandler {
	return &handler{
		s: services.New(),
	}
}

func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request)) {
	w.Write(h.s.Health())
}
```

MohammadAlhallaq:
> Added a few comments - tested out the PR everything seems to work however, the DB doesnt actually get hooked up. Wondering if there is a way we can fully test out this feature and implement it for a user to use

I will tackle those asap, and try to figure out a way to make the service accessible via a route for testing

Melkeydev:
> > Added a few comments - tested out the PR everything seems to work however, the DB doesnt actually get hooked up. Wondering if there is a way we can fully test out this feature and implement it for a user to use
> 
> I will tackle those asap, and try to figure out a way to make the service accessible via a route for testing

Ping me when this is ready for a another review :)

MohammadAlhallaq:
Working on creating a route to perform database health checks.

/*
  Checklist:
  - [x] chi
  - [x] echo
  - [ ] fiber
  - [ ] gorilla
  - [ ] routerRoutes
  - [ ] std package
*/

cmd/template/DBDriver/postgres.go
MohammadAlhallaq: I would appreciate it if you could take a look now. I wonder if this could be implemented in a more suitable way?

cmd/template/dbdriver/postgres.go
```diff
@@ -0,0 +1,50 @@
...
+type Service interface {
+	Health() map[string]string
+}
+
+type service struct {
+	db *sql.DB
+}
+
+func New() *service {
+	connStr := "user=pqgotest dbname=pqgotest sslmode=verify-full"
```
Melkeydev, line 27: I am going to add a general comment that will impact all of the DB implementations:
- using as it, curling localhost:8080/health fails due to connection not existing.
I am curious - should we include .env variables so that a user can update them and actually have a connection working OOB?
Because right now, if we include this feature and it doesnt work I feel a user can have poor experience.
If we allow them to update the .env (somehow?) and use a tool like godotenv, we can populate it and allow a proper connection.
And maybe (just a suggestion)if there is no .env we can just skip having the health check route implemented?
What are your thoughts?

Melkeydev (changes requested):
I have an idea as well:
- Since the goal is to have something working right away, what do you think of adding a docker-compose file that spins up the DB they choose? that way we can control the .env vars right away and a user can have something working ASAP.
If they want to change the connection strings, they are free to do so.

I actually think this is a good idea, here is an example for postgres:
```
version: "3.1"

services:
  db:
    container_name: "lowball"
    image: postgres:12.4-alpine
    volumes:
      - "./database/postgres-data:/var/lib/postgresql/data:rw"
    ports:
      - "5432:5432"
    environment:
      POSTGRES_DB: "postgres"
      POSTGRES_USER: "postgres"
      POSTGRES_PASSWORD: "postgres"
    restart: unless-stopped
    ```

File: cmd/template/dbdriver/mongo.go [hunk 11]
```diff
@@ -0,0 +1,52 @@
...
+)
+
+type Service interface {
+	Health() map[string]string
+}
+
+type service struct {
+	db *mongo.Client
+}
+
```
Melkeydev, line 25: nit:
maybe we include this in the existing Server struct?

MohammadAlhallaq: I've just tried this approach, and it is possible to implement, although it would require us to change the server template by making probably more versions of it depending on the chosen DB driver, while currently, we are only depending on the database package and making some additions to the routes template when the user chooses a driver to get installed, while the server package remains untouched.

I would like to know your take on this. If you think I might have missed your point,

cmd/template/dbdriver/postgres.go
MohammadAlhallaq: In my opinion, having the environment env available for the user to set makes more sense than disabling the route for the health check. because disabling the route could confuse the user more than showing an error log indicating that the connection fails due to the credentials not being set correctly.

cmd/template/dbdriver/mongo.go
MohammadAlhallaq: you mean something like this 

```go
package server

type Server struct {
	port int
	database.Service
}


And then use it directly in the Server template to create a new instance and call the HealthHandler method?

Comment (MohammadAlhallaq) [reply cmd/template/dbdriver/mongo.go:25]: Hmm, that's right. I will try to tweak the implementation tonight.

Comment (MohammadAlhallaq) [reply cmd/template/dbdriver/mongo.go:25]: something like this maybe ? 

```go

type Server struct {
	port int
	db   *database.Service
}


func NewServer() *http.Server {

	NewServer := &Server{
		port: port,
		db:   database.New(),
	}

	// Declare Server config
	server := &http.Server{
		Addr:         fmt.Sprintf(":%d", NewServer.port),
		Handler:      NewServer.RegisterRoutes(),
		IdleTimeout:  time.Minute,
		ReadTimeout:  10 * time.Second,
		WriteTimeout: 30 * time.Second,
	}

	return server
}
```

then using it in the route package like this 

```go
func (s *Server) RegisterRoutes() http.Handler {
	r := gin.Default()
	r.GET("/", s.helloWorldHandler)
	r.GET("/health", s.healthHandler)

	return r
}


func (s *Server) healthHandler(c *gin.Context) {
	c.JSON(http.StatusOK, s.db.Health())
}
```

cmd/template/dbdriver/postgres.go
Melkeydev: Good point on not bloating this PR - I am going to add an issue to include the Docker feature (https://github.com/Melkeydev/go-blueprint/issues/65)

However - my current question is that if someone spins up the Repo as is and they try to query the routes, it will fail out. 
isn't this a bit of a concern for a feature we are adding?
BTW I was also thinking due to the size of this PR, this needs to be a separate Major release at this point.

cmd/template/dbdriver/mongo.go
Melkeydev: Yeah - this is exactly what I was thinking of as well.
However, like in my previous comment, if we do not have an actual DB spun up for the user then
```r.GET("/health", s.healthHandler)```
Will error out if they try to cURL if after running `make run`

File: cmd/template/dbdriver/postgres.go (hunk 10 above)
Comment (MohammadAlhallaq) [reply cmd/template/dbdriver/postgres.go]: After adding the env file and granting users the option to specify their database credentials before accessing the database health check, it is essential for the user to include the credentials in their env file even after incorporating the docker starter files. There seems to be no way around this requirement, as users must edit the env file to ensure proper functionality of the database.

similar to other frameworks like Laravel, users are required to generate the env file and provide the correct credentials to ensure that the started API route is working properly

File: cmd/template/dbdriver/mongo.go (hunk 11 above)
Comment (MohammadAlhallaq) [reply cmd/template/dbdriver/mongo.go:27]: Will work on the new implementation for all the drivers asap? Also, I've answered your question in the earlier convo

File: cmd/template/dbdriver/postgres.go (hunk 10 above)
Comment (MohammadAlhallaq) [reply cmd/template/dbdriver/postgres.go]: Also, about your question, I was thinking that in addition to adding the env file whenever the user chooses a database driver, we might also include a starter migration file. This could serve as a way to prompt the user to add their database credentials to the env file. 
What are your thoughts on this? Is it worth to tackle it in this PR, or should we add it in a separate PR to expedite the current one and get it merged asap?
@Melkeydev

**Comment by Melkeydev** [comment]
@MohammadAlhallaq Do you have Discord? Can you join https://discord.gg/melkeydevhouse ?

**Comment by MohammadAlhallaq** [comment]
> @MohammadAlhallaq Do you have Discord? Can you join https://discord.gg/melkeydevhouse ?

yup, will join!

File: cmd/program/program.go [hunk 12]
```diff
@@ -196,16 +272,22 @@ func (p *Project) CreateMainFile() error {
 		return err
 	}
 
-	err = p.CreateFileWithInjection(internalServerPath, projectPath, "server.go", "server")
+	if p.DBDriver != "none" {
+		err = p.CreateFileWithInjection(internalServerPath, projectPath, "routes.go", "routesWithDB")
+		err = p.CreateFileWithInjection(internalServerPath, projectPath, "server.go", "serverWithDB")
+	} else {
+		err = p.CreateFileWithInjection(internalServerPath, projectPath, "routes.go", "routes")
+		err = p.CreateFileWithInjection(internalServerPath, projectPath, "server.go", "server")
```
20pa5a1210, line 280: i think you should error every time

MohammadAlhallaq: Yeah, I'm not totally convinced with the whole approach. I'm probably going to change it to a more appropriate way.

.idea/go-blueprint.iml
mimatache, line 1: this need to be deleted and `.idea` need to be added to .gitignore .

cmd/template/framework/files/routes/gorilla.go.tmpl
```diff
@@ -25,5 +25,5 @@ func (s *Server) helloWorldHandler(w http.ResponseWriter, r *http.Request) {
 		log.Fatalf("error handling JSON marshal. Err: %v", err)
 	}
 
-	_, _ = w.Write(jsonResp)
-}
+	w.Write(jsonResp)
```
mimatache, line 28: This will cause the linter to fail

BKR-dev (approved):
LGTB

Melkeydev (approved):
Looks
Fucking
Incredible
To Me
My honor to MERGE THIS

MohammadAlhallaq:
> Looks Fucking Incredible To Me My honor to MERGE THIS

thanks man, I've really enjoyed working on it.

PR #24, opened 2023-11-04
A github action to create a release on git tag

MitchellBerend:
Added a workflow that builds binaries for arm and x86 and creates a release with them on a git tag with the pattern `v*.*.*`

You can check out [my fork](https://github.com/MitchellBerend/go-blueprint) for an example with `v69.420.0`

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -12,7 +12,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Set up Go
         uses: actions/setup-go@v4
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,0 +1,40 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+
+permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,40 @@
...
+      - "v*.*.*"
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
itschip, line 15: You can use v4

.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.21.1'
```
itschip, line 20: The version that is used in the projet is 1.20, if that will have some side-effects(?)

MitchellBerend: I thought go had a full backwards compatible compiler, I can change this back though.

itschip: Nah, ur probably right on this one! :)

itschip:
LGTM

Melkeydev (approved):
LGTM

PR #27, opened 2023-11-05
A concept implementation of a websocket endpoint

### LINKED ISSUE #25: Websocket example implementation
Would a ping pong web socket endpoint example add a lot of value? The way I imagine it, it would add a [dependency](https://pkg.go.dev/nhooyr.io/websocket) to the generated project, this dependency does get linked to by the [docs](https://pkg.go.dev/golang.org/x/net/websocket).

I think gorilla has their own websocket implementation but I haven't worked with that one yet.

naive example implementation
```go

handler(w http.ResponseWriter, req *http.Request) {
	socket, _ := websocket.Accept(w, req, nil)
	ctx := req.Context()
	for {
		_, socketBytes, _ := socket.Read(ctx)
		if string(socketBytes) == "PING" {
			_ = socket.Write(ctx, websocket.MessageText, "PONG")
		} else {
			_ = socket.Write(ctx, websocket.MessageText, "HUH?")
		}
	}
}


```

MitchellBerend:
Resolves #25

This pr adds a websocket endpoint for all frameworks.

- [x] chi
- [x] echo
- [x] fiber
- [x] gin
- [x] gorilla
	~~- [ ] decide on websocket implementation~~
- [x] httpRouter
- [x] std lib
- [x] Decide on error logging
- [ ] Added error [outputs](https://github.com/Melkeydev/go-blueprint/pull/27/files/02b3dd8f9c015b08ea2c5f6c09a362fdec2cf496#diff-82f92bef562cb03ce68889e141bf547900defec856e3f6a5a1cbbeca8f090c88)

### DIFF
--- a/cmd/template/chiRoutes.go
+++ b/cmd/template/chiRoutes.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
+	"nhooyr.io/websocket"
 )
 
 func (s *Server) RegisterRoutes() http.Handler {
@@ -35,6 +36,7 @@ func (s *Server) RegisterRoutes() http.Handler {
 	r.Use(middleware.Logger)
 
 	r.Get("/", s.helloWorldHandler)
+	r.Get("/ws", s.pingPongWebsocketHandler)
 
 	return r
 }
@@ -51,5 +53,17 @@ func (s *Server) helloWorldHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write(jsonResp)
 }
 
+func (s *Server) pingPongWebsocketHandler(w http.ResponseWriter, r *http.Request) {
+	socket, _ := websocket.Accept(w, r, nil)
+	ctx := r.Context()
+	for {
+		_, socketBytes, _ := socket.Read(ctx)
+		if string(socketBytes) == "PING" {
+			_ = socket.Write(ctx, websocket.MessageText, []byte("PONG"))
+		} else {
+			_ = socket.Write(ctx, websocket.MessageText, []byte("HUH?"))
+		}
+	}
+}
 `)
 }
--- a/cmd/template/chiRoutes.go
+++ b/cmd/template/chiRoutes.go
@@ -21,7 +21,8 @@ func (c ChiTemplates) Routes() []byte {
 func MakeChiRoutes() []byte {
 	return []byte(`package server
 
-import ( "encoding/json"
+import (
+	"encoding/json"
 	"log"
 	"net/http"
 
--- a/cmd/template/chiRoutes.go
+++ b/cmd/template/chiRoutes.go
@@ -21,8 +21,7 @@ func (c ChiTemplates) Routes() []byte {
 func MakeChiRoutes() []byte {
 	return []byte(`package server
 
-import (
-	"encoding/json"
+import ( "encoding/json"
 	"log"
 	"net/http"
 
@@ -58,6 +57,9 @@ func (s *Server) pingPongWebsocketHandler(w http.ResponseWriter, r *http.Request
 
 	if err != nil {
 		log.Print("could not open websocket")
+		// pray this works for your user
+		_, _ = w.Write([]byte("could not open websocket"))
+		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 
--- a/cmd/template/chiRoutes.go
+++ b/cmd/template/chiRoutes.go
@@ -1,5 +1,7 @@
 package template
 
+import "fmt"
+
 // ChiTemplates contains the methods used for building
 // an app that uses [github.com/go-chi/chi]
 type ChiTemplates struct{}
@@ -19,7 +21,7 @@ func (c ChiTemplates) Routes() []byte {
 // MakeChiRoutes returns a byte slice that represents 
 // the internal/server/routes.go file when using Chi.
 func MakeChiRoutes() []byte {
-	return []byte(`package server
+	return []byte(fmt.Sprintf(`package server
 
 import (
 	"encoding/json"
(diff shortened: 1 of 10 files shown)


MitchellBerend:
The current implementation does not have any error handling at all, which is obviously not good. I think either a 5xx should be returned or it should be **very** clear that errors are not handled at all.

Should I just return a 5xx when some error is encountered?

MitchellBerend:
The current implementation does not have any error handling at all, which is obviously not good. I think returning a 5xx is the best option but it might lead to users not looking at the implementation since it "works" as is.

As this is still a wip I have 2 questions as for the direction of this pr

Should this handler just return a 5xx when any error is encountered?
Should there be a comment about specifically looking at the error handling in this endpoint?

I can see a point for doing both but I'd like your opinion on that too before I keep working on this.

MitchellBerend:
Gorilla/mux has their [own](https://github.com/gorilla/websocket) websocket implementation which could be used for the generated code.  Something that has to be decided is if gorilla/mux is also going to use a different websocket implementation than the rest. Apart from fiber which also uses it's own implementation, all the other implementations use the same websocket library. I think an argument could be made for keeping the generated code in the same ecosystem as much as possible.

Let me know what you think.

MitchellBerend:
I slept on the error logging decision and I decided that it should not be up to the user to implement the error handling for the generated code. I had some trouble getting the fiber websocket endpoint implementation working yesterday and adding error logging there helped a lot.

MitchellBerend:
Okay I decided to push back the possible [gorilla/websocket](https://github.com/gorilla/websocket) implementation since this pr has gotten quite big already and I don't want to keep expanding it's scope.

Melkeydev:
This looks good.
I am going to test it out right now and check out the code.
I will give my opinion on the error handling now too.
But there is also a merge conflict that needs to be resolved

cmd/template/chiRoutes.go
```diff
@@ -47,5 +49,38 @@ func (s *Server) helloWorldHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write(jsonResp)
 }
 
+func (s *Server) pingPongWebsocketHandler(w http.ResponseWriter, r *http.Request) {
+	socket, err := websocket.Accept(w, r, nil)
+
+	if err != nil {
+		log.Print("could not open websocket")
+		// pray this works for your user
```
Melkeydev, line 57: ?? lol

cmd/template/chiRoutes.go
```diff
@@ -47,5 +49,38 @@ func (s *Server) helloWorldHandler(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+
+		if string(socketBytes) == "PING" {
+			if err := socket.Write(ctx, msgType, []byte("PONG")); err != nil {
+				log.Print("could not write to socket")
+				return
+			}
+		} else {
+			if err := socket.Write(ctx, msgType, []byte("HUH?")); err != nil {
```
Melkeydev, line 78: nit: the HUH is kind of funny but i am wondering if we can send a better response lol
Maybe:
"This is another message not PING"

cmd/template/echoRoutes.go
```diff
@@ -39,5 +44,40 @@ func (s *Server) helloWorldHandler(c echo.Context) error {
...
 	return c.JSON(http.StatusOK, resp)
 }
+
+func (s *Server) pingPongWebsocketHandler(c echo.Context) error {
+	writer := c.Response().Writer
+	socket, err := websocket.Accept(writer, c.Request(), nil)
+
+	if err != nil {
+		log.Print("could not open websocket")
+		// pray this works for your user
```
Melkeydev, line 54: x_x

cmd/template/httpRoutes.go
```diff
@@ -77,5 +80,39 @@ func (s *Server) handler(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+		} else {
+			if err := socket.Write(ctx, msgType, []byte("HUH?")); err != nil {
+				log.Print("could not write to socket")
+				return
+			}
+		}
+	}
+}
```
Melkeydev, line 116: i just thought of something:
since this is basically copy and pasted from one to the other, it makes more sense just to create one template file that contains the logic for all of them 
For example: https://github.com/Melkeydev/go-blueprint/blob/main/cmd/template/main.go#L7 
is used across the majority of the set up except the fiber one

cmd/template/routerRoutes.go
```diff
@@ -42,6 +44,38 @@ func (s *Server) helloWorldHandler(w http.ResponseWriter, r *http.Request) {
...
+			log.Print("could not read from websocket")
+			return
+		}
+
+		if string(socketBytes) == "PING" {
+			if err := socket.Write(ctx, msgType, []byte("PONG")); err != nil {
+				log.Print("could not write to socket")
+				return
+			}
+		} else {
```
Melkeydev, line 72: Also - i am wondering if this else is even needed?

Melkeydev (changes requested):
This looks great.
I think the error handling is good. I left some nits on the review but I do think that instead of copy and pasting the code across, could we possibly abstract it into a dedicated file and then use that in the routes? LMK

cmd/template/chiRoutes.go
MitchellBerend: I think it's odd to handle error cases in the error handling block so I just ignored the potential error returned in the line under that and put a comment there indicating a potential issue but I can just remove this comment.

MitchellBerend:
I think all of the points are addressed but I still need to test and format the code properly since I can't do that on this computer.

- [ ] run gofmt
- [ ] test functionality

itschip:
We might also want to send a close message to gracefully shutdown the websocket

MitchellBerend:
> We might also want to send a close message and gracefully shutdown the websocket.

This sounds like a good idea tbh. If this PR gets accepted I think a new PR for that would be appropriate since this PR is large enough as it is imo.

Melkeydev (changes requested):
When i run go build:
```
❯ go build
# github.com/melkeydev/go-blueprint/cmd/template
cmd/template/main.go:66:2: syntax error: unexpected newline in argument list; possibly missing comma or )

```

MitchellBerend:
Oh sorry I forgot to test this at home...

cmd/template/echoRoutes.go
```diff
@@ -43,5 +48,43 @@ func (s *Server) helloWorldHandler(c echo.Context) error {
...
 }
+
+func (s *Server) pingPongWebsocketHandler(c echo.Context) error {
+	w := c.Response().Writer
+	r := c.Request()
+	errorMessage := []byte("This is another message not PING")
+	socket, err := websocket.Accept(w, r, nil)
+
+	if err != nil {
+		log.Print("could not open websocket")
```
smantic, line 59: We're eating the error here and not printing it anywhere

cmd/template/main.go
```diff
@@ -23,6 +23,50 @@ func main() {
...
+// handler. This function expects some variables set in the template. See echo
+// router for more details.
+//
+// w http.ResponseWriter
+// r *http.Request
+func websocketTemplate() []byte {
+	return []byte(`
+	errorMessage := []byte("This is another message not PING")
+	socket, err := websocket.Accept(w, r, nil)
+	defer socket.Close(websocket.StatusGoingAway, "server closing websocket")
```
itschip, line 36: Shouldn't this be moved after the err check for socket.Accept?

itschip: Moving the socket.Close that is

smantic: You'll want to check the error first, then defer the close. Otherwise you will get what we saw a stream. A Panic because we tried to close a nil socket.

PR #28, opened 2023-11-05
Added a linting job that triggers on PR events

MitchellBerend:


MitchellBerend:
Maybe you should also add branch protections that requires this job to pass so you don't shoot yourself in the foot as well. There is also a setting in there somewhere that only allows squash merges so the git history doesn't look like shit.

cmd/program/program.go
```diff
@@ -46,8 +46,11 @@ var (
 func (p *Project) ExitCLI(tprogram *tea.Program) {
 	if p.Exit {
 		// logo render here
-		tprogram.ReleaseTerminal()
-		os.Exit(1)
+		if err := tprogram.ReleaseTerminal(); err != nil {
+			log.Fatal(err)
+		} else {
+			os.Exit(1)
+		}
```
Melkeydev, line 53: I dont think this if /else is necessary

.github/workflows/linting.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.21.1'
+
+      - name: Install dependencies
+        run: go mod tidy
+
+      - name: Install dependencies
+        run: go mod download
```
Melkeydev, line 21: Two install dependencies steps?

.github/workflows/linting.yml
```diff
@@ -0,0 +1,25 @@
+name: Linting
```
Melkeydev, line 1: I am wondering if this can be a pre-commit step? Just asking if this is necessary step to do here

Melkeydev (changes requested):
Wondering about the execution of this Job and if we really need it

MitchellBerend:
This job isn't strictly necessary as a precommit script can do the same, but that could be skipped or forgotten. It's on the maintainer to install the script, I think having a safety net for that makes sense.

Melkeydev (approved):
LGTM

Issue #34, opened 2023-11-06
Add Air for hot reloading


Melkeydev:
We should add https://github.com/cosmtrek/air
to the app so that we get hot reloading

PR #35, opened 2023-11-06
Feature: air toml file for air hot reload

20pa5a1210:
Resolving Issue #34


20pa5a1210:
iam injecting `.air.toml` file directly instead doing `air init` because default toml file looking for `entry point` at root but our project root file  is under cmd folder so iam injecting `.air.toml` so we can directly run `Make run` command from Makefile inside  `.air.toml`

and i made `make watch` command in makefile to run air or we can use `air` directly either way both are working

20pa5a1210:
so i have small doubt if user doesn't have air installed globally it doesn't work  if we run `air` command  right? 
image

can we install air package like this instead if user does not have air globally ?

Melkeydev (commented):
This looks great.
My opinion is that we should not be installing things on a users machine if they dont want it.
So what i think could make sense is if they dont have it we can prompt to install it themselves or something.
So as it stands, I think this PR is G2G
But I am going to test it out now

Melkeydev (approved):
Tested - LGTM

PR #44, opened 2023-11-08
Fix/air: Air Install Globally 

20pa5a1210:
###  running small bash script if user don't have Air installed globally it asks Y/N to install the AIR #

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -223,6 +223,13 @@ func (p *Project) CreateMainFile() error {
         return err
     }
 
+    // go install air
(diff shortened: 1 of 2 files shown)


20pa5a1210:
# DEMO
[airbashtest.webm](https://github.com/Melkeydev/go-blueprint/assets/95855318/07f5319a-c66f-43fa-81c4-64620e8f4475)

dreulavelle:
Would love to get this approved. Good feature to have for a project like this I think 🤔

Melkeydev (approved):
LGTM

PR #48, opened 2023-11-10
Update README.md with New logo

chandez:
The PSD file is also included. Feel free to use it however you like.
Thanks to @mikelerch for your previous logo.

ヽ(•‿•)ノ


Melkeydev:
@chandez Thank you for the PR - however with something like a Logo this is the complete identity of the project.
I am going to close the PR because the logo here is just not what I have for the vision of the project right now!

PR #51, opened 2023-11-10
feat: move templates to embedded files

mimatache:
fixes #49

PR #53, opened 2023-11-10
implemented .env varibale to be load

stgonzales:
hey, I've implemented `.env` file load for `fiber` and `http` frameworks
please let me know if there's something that could be better done.
tks


cmd/utils/utils.go
```diff
@@ -57,3 +57,14 @@ func GoFmt(appDir string) error {
...
 }
+
+// GoModTidy runs "go mod tidy" in a selected directory
+func GoModTidy(appDir string) error {
+	if err := ExecuteCmd("go",
+		[]string{"mod", "tidy"},
+		appDir); err != nil {
+		return err
+	}
+
```
basokant, line 68:
```suggestion
	err := ExecuteCmd("go",
		[]string{"mod", "tidy"},
		appDir);

	if err != nil {
		return err
	}
```

nit change. should probably also just run go fmt

stgonzales: Im following the format used in previous cmds, maybe would be good to stick with same format until have some standards (?)

cmd/template/main.go
```diff
@@ -113,6 +113,14 @@ tmp_dir = "tmp"
...
 }
 
+func EnvFileTemplate() []byte {
+  return []byte(`#Environment Variables
+  
+# Port to run the application on
+PORT="8080"
+  `)
+}
+
```
Melkeydev, line 123: This is great! However, we have another [PR](https://github.com/Melkeydev/go-blueprint/pull/51)
That is moving go-blueprint away from this style and using tmpl + embed more.
Can we make that change here as well?
Also, we should add a .env ignore in the .gitignore generator as well in this PR!

Melkeydev (changes requested):

Melkeydev:
Also there is some conflicts that need to be resolved

cmd/template/main.go
stgonzales: good call, I've re-based and implemented as embed, also add `.env` to gitignore

PR #59, opened 2023-11-11
Print the Equivalent Non-interactive Command

### LINKED ISSUE #56: Print the Equivalent Non-interactive Command
After completing the last step of the interactive CLI, the user should be able to easily copy and paste the equivalent shorter non-interactive go-blueprint command with the proper flags and program arguments set.

For example, after the user selects Gin in the interactive UI, the output at the end could be something like this:

```
Tip: Repeat the equivalent Blueprint with the following non-interactive command
go-blueprint create --name my-project --framework gin
```

Currently, this is not terribly useful, BUT as we add more steps it could save a lot of time.

basokant:
closes #56 

- Prints the equivalent non-interactive command with flags and values
- Only does this when in interactive mode (when no flags are set by the user)

image

Demo:

https://github.com/Melkeydev/go-blueprint/assets/70033855/960ffedb-4fb2-4fd0-9587-1b52373505f0

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -55,6 +55,8 @@ var createCmd = &cobra.Command{
 			ProjectName: &textinput.Output{},
 		}
 
+		isInteractive := !utils.HasChangedFlag(cmd.Flags())
+
 		flagName := cmd.Flag("name").Value.String()
 		flagFramework := cmd.Flag("framework").Value.String()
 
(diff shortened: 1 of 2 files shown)


basokant:
Let me know what you think of this tiny lil feature I added @Melkeydev, I think it might be useful in the future, tried to make it as future proof as possible.

Also, I'm not sure if I like the message, so feel free to suggest a change, or change the message directly yourself 👍

Melkeydev (approved):
LGTM

basokant:
W rizz

PR #64, opened 2023-11-11
Gives an error if a directory already exists. Colorized the errors when doing `create` if a directory already exists and not empty

Jamlie:
I made the project check the availability the given directory, if it does exist and it's not empty, it gives an error and exits.

The error messages when creating a project - if occurred - are now colored with #FF8700

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -1,18 +1,17 @@
 package cmd
 
 import (
-	"fmt"
-	"log"
-	"os"
-	"strings"
+        "fmt"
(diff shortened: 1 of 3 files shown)


Melkeydev:
@Jamlie I think there is a massive format issue going on here which makes it tough to see what are your specific changes
Also, there are some merge conflicts :/

Jamlie:
yea i think it's from tabs and spaces, i'll fix it asap

PR #67, opened 2023-11-11
Feat/docker compose dbs

muandane:
add docker compose file


Melkeydev:
This is really cool - however what we should do instead is have these set as templ + embed files and when the user from [this PR](https://github.com/Melkeydev/go-blueprint/pull/22)
chooses what DB they want, then we can spin up that specific one in their generated project

PR #69, opened 2023-11-11
GitHub action, Jenkins & Tests

Ujstor:
Hi, I added CI/CD for Jenkins and GitHub Actions, a simple template that can be updated. Also, I created tests after the framework is spun up. I was paying attention to the DB PR to ensure that, after merging, I don't create unnecessary conflicts.

This is my first PR, and I hope it helps.


Melkeydev (changes requested):
This is an incredible PR however i dont think we need to have an entire Jenkins pipeline.
As of right now, our CI/CD is handled by GHA and it works fine.

PR #77, opened 2023-11-12
Adding test and some refactor

itschip:
**Features:**
- Test for checking missing values between allowed types and the framework map.

**Refactoring**
- Moving some logic out of the `Run` command to "clean it up".
- New file `project.go` which inits and returns the structs needed to create the project. Would likey come in handy should there be other commands needing to use the project as well.
- Some small changes in the ui/text after creating the project.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -98,7 +98,11 @@ func createNamePrompt(cmd *cobra.Command, project *program.Project, options *ste
 	project.ExitCLI(tprogram)
 
 	project.ProjectName = options.ProjectName.Output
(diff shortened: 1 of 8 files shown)


itschip:
We could probably actually just read straight from the framework map instead of using another slice for allowed types

Melkeydev:
@itschip Is this still in draft? There are also conflicting files here as well

itschip:
> @itschip Is this still in draft? There are also conflicting files here as well

I'll clean it up and see what's changed and set it up for review within the weekend :)

PR #78, opened 2023-11-12
Fix: Print Non Interactive Command

### LINKED ISSUE #73: Bug: The non-interactive command from tip message is broken
Bug introduced in #59 (I messed up team).
Remind me not to code at 4am. Also we very much need unit tests.

- [x] include `create` in command
- [x] fix `standard library` framework option to be `standard-library` instead.

image

basokant:
closes #73

Confirmed that non interactive command in program output is correct.

- [x] include `create` in command
- [ ] fix `standard library` framework option to be `standard-library` instead.
- [x] move name flag to the end of the output so that it can easily be replaced

image


Melkeydev:
@basokant You need to rebase this to pass the linting.
Also - is this still a draft?
Also, there are some conflicts

PR #79, opened 2023-11-12
Add abort if directory already exists

### LINKED ISSUE #54: Abort if project directory already exists
## Description

The application currently creates a new project directory and files based on the provided project name, even if a directory with the same name already exists. This behavior is problematic as it could lead to unintended overwriting of existing project data. The application should instead abort creation when a project with the same name already exists.

## Steps to Reproduce

Run `go-blueprint create` in a directory with a sub directory called the same as the name of the project. 

## Expected Result

When attempting to create a project with a name matching an existing project, the application should provide an error message and abort the creation process to avoid overwriting the existing project.

nickwarters:
fixes #54 

Aborts if project directory already exists

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -3,11 +3,13 @@
 package program
 
 import (
+	"errors"
 	"fmt"
 	"html/template"
 	"log"
 	"os"
 	"strings"
+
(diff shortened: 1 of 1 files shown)


Melkeydev:
Another PR was implemented that does this! (it was submitted first)

PR #80, opened 2023-11-12
Framework flag shell completion

### LINKED ISSUE #61: Proposal: shell completion for `--framework` flag
It would be nice if the `completion` command would also generate completion for the `--framework` flag. It will help with feature discovery for new users.

MitchellBerend:
Resolves #61 

This pr adds shell completion for the --framework flag. This feature should
help users with discovering framework options, prevent typos and last but not
leas, typing less.

Besides the tab completion it also replaces the `string` types for the
`FrameworkType` with a type alias `Framework`. The `VarP` function registers
a flag that always returns a valid string which then gets turned into a specific
type to make it type safe.

cc @basokant


basokant (approved):
Lgtm! I'm guessing u still need to run the completion command and add the completion script?

MitchellBerend:
> Lgtm! I'm guessing u still need to run the completion command and add the completion script?

Yes, you need to register the flag for the completion generation with a function that gets called for the option but that's a simple one line function. 

https://github.com/MitchellBerend/go-blueprint/blob/7bef3e027fa9491d4c8a25ae1f40ff8124581bed/cmd/create.go#L47-L49

https://github.com/MitchellBerend/go-blueprint/blob/7bef3e027fa9491d4c8a25ae1f40ff8124581bed/cmd/frameworks/frameworkFlag.go#L42-L44

basokant:
> > Lgtm! I'm guessing u still need to run the completion command and add the completion script?
> 
> 
> 
> Yes, you need to register the flag for the completion generation with a function that gets called for the option but that's a simple one line function. 
> 
> 
> 
> https://github.com/MitchellBerend/go-blueprint/blob/7bef3e027fa9491d4c8a25ae1f40ff8124581bed/cmd/create.go#L47-L49
> 
> 
> 
> https://github.com/MitchellBerend/go-blueprint/blob/7bef3e027fa9491d4c8a25ae1f40ff8124581bed/cmd/frameworks/frameworkFlag.go#L42-L44
> 
> 
> 
> cc @basokant 

This is awesome! I thought you would have to add the shell completion script that's generated from 'go-blueprint completion', but this is way better!

Melkeydev (changes requested):
i think this could be out of scope for Go-blueprint because modifying dot files seems to be a little much for this CLI tool.
Personally, I wouldnt want something I installed just for spinning up Go boiler plate to touch my dot files.
I think certain users can have a poor experience with that.

Issue #58, opened 2023-11-11
Proposal: Add optional default root directory


nickwarters:
If you have all of your projects in a specific directory, rather than needing to run the command in that directory, it could be useful to be able to run the command from anywhere and the project be created in your base/root project directory.

For example I could be in `~/some/random/path` and by running `go-blueprint create --name new-project` would create the project in `~/dev/new-project`

basokant:
Not sure about the default root directory, but I do agree that only being able to choose the name is a bit limiting. It could be nice to instead have the step be called "path" which would allow you to pass in an absolute/relative path to the directory that the project should live. if the directory doesn't exist, it would be nice if it created it (same with any of it's parent directories).

@nickwarters Would this give the same utility? I've seen other bootstrap CLI tools do this too. (ex. `npm create astro@latest`)

nickwarters:
I think a path option would be good yea. 

My original idea was kinda have it optional rather than default. 
Be able to set it in a global config or something.

nickwarters:
@basokant were you going to take a crack at this? If not I'd like to :)

PR #82, opened 2023-11-12
Add --path flag

nickwarters:
Adds a `--path` flag following suggestion in comments of #58 

- use `.` for current directory
- use relative path to current directory e.g. `./test-folder`
- paths without `/`, `\` or `.` prefix will be treated as relative to current directory 
- use absolute paths e.g. `/users/{username}/dev`
- `~` is expanded to `$HOME` env variable so `~/dev` and `/Users/{username}/dev` are equivalent if `$HOME` is set to `/Users/{username}`


nickwarters:
@Melkeydev what are your thoughts on this feature? Worthwhile or not? If not, let's close off and knock another one off the queue :)

cmd/create.go
```diff
@@ -92,6 +95,16 @@ var createCmd = &cobra.Command{
 			}
 		}
 
+		if flagPath == "" {
+			tprogram = tea.NewProgram(textinput.InitialTextInputModel(options.ProjectPath, "Where should the project be created? (use '.' for current directory)", project))
```
Melkeydev, line 99: TBH Idk if i like forcing a user to add a path to explicitly state where the project should be created.
Also, where would the option be in the CLI?

Melkeydev (changes requested):

PR #88, opened 2023-11-12
Linting before approval

CamPen21:
# Motivation

Running this workflow after approval causes reprocessing as it may or may not break. If it does break unfortunately changes need to happen and a new review needs to be done.

# Changes

- The workflow runs on a push, this will run more frequently but I have added caching to deps to make it fast enough so it's not annoyingly slow.


Melkeydev (approved):
LGTM

PR #89, opened 2023-11-12
Typesafety for `framework` and `driver` flags

MitchellBerend:

## Problem/Feature

Currently the framework and driver flags is passed as a string that can contain any value. The value of this flag is then manually validated and compared to hard coded values.

## Description of Changes: 

Converted the framework driver flags to use a struct to ensure the content is valid throughout the code base. This creates a single place to add or change framework options. It also adds constants to compare values against.

This is a continuation of #61, this pr adds a type alias `Framework` that only allows for certain values to be passed as a flag.

## Checklist

~~- [ ] I have updated the documentation (if applicable)~~ (Not applicable)


cmd/flags/frameworks.go
```diff
@@ -0,0 +1,48 @@
...
+
+var AllowedProjectTypes = []string{string(Chi), string(Gin), string(Fiber), string(GorillaMux), string(HttpRouter), string(StandardLibrary), string(Echo)}
+
+// This interface is required on a type to make it useable as a flag
+//
+//	type Value interface {
+//		String() string
+//		Set(string) error
+//		Type() string
+//	}
```
Melkeydev, line 31: Leftover?

cmd/flags/frameworks.go
```diff
@@ -0,0 +1,48 @@
...
+	Chi             Framework = "chi"
+	Gin             Framework = "gin"
+	Fiber           Framework = "fiber"
+	GorillaMux      Framework = "gorilla/mux"
+	HttpRouter      Framework = "httprouter"
+	StandardLibrary Framework = "standard-library"
+	Echo            Framework = "echo"
+)
+
+var AllowedProjectTypes = []string{string(Chi), string(Gin), string(Fiber), string(GorillaMux), string(HttpRouter), string(StandardLibrary), string(Echo)}
```
Melkeydev, line 23: Nit:
To update this list if someone wants to add a new framework they would need to update this in both spots..
Is there a way we can get around this?
Follow - up: is it only these two places or more?

Melkeydev (changes requested):

cmd/flags/frameworks.go
MitchellBerend: Oh woops

cmd/flags/frameworks.go
MitchellBerend: > Is there a way we can get around this?

I could try and do something with `iota` and build the slice dynamically I think.


> Follow - up: is it only these two places or more?

You would also need to add the actual implementation logic but besides that these are the only places.

MitchellBerend: I did a little digging and all different implementations I found either also had 2 spots to maintain or weren't type safe

cmd/create.go
```diff
@@ -66,23 +68,11 @@ var createCmd = &cobra.Command{
...
-		if flagDBDriver != "" {
-			isValid := isValidDBDriver(flagDBDriver, allowedDBDrivers)
-			if !isValid {
-				cobra.CheckErr(fmt.Errorf("database driver '%s' is not valid. Valid types are: %s", flagDBDriver, strings.Join(allowedDBDrivers, ", ")))
-			}
-		}
+		// VarP already validates the contents of the framework flag.
+		// If this flag is filled, it is always valid
+		flagFramework := flags.Framework(cmd.Flag("framework").Value.String())
+		flagDBDriver := flags.Database(cmd.Flag("driver").Value.String())
```
Melkeydev, line 75: How does error handling on this work then?

Melkeydev (changes requested):

cmd/create.go
MitchellBerend: The validations happen by `VarP` that function registers the flags and only allows for valid values to pass through. If there is an invalid value, `VarP` prints the last arg to the terminal and exists the program. An empty string is passed if the flag is not used on the command line. I think technically you can also pass an empty string (`go-blueprint -f "") but I don't think we should prevent this since it is already handled the same as not using the flag at all.


https://github.com/MitchellBerend/go-blueprint/blob/e443e0d9a346b69448f1d46419c0048418998dd0/cmd/create.go#L45-L46

Melkeydev (approved):
LGTM

PR #90, opened 2023-11-12
Feature/caddy

rustafariandev:
Create a template of a caddy server


Melkeydev:
I am wondering if caddy is actually needed here for Go-blueprint?
The existing frameworks/options creates a basic RESTFUL API with a http server.
I dont think Caddy is appropriate here?

rustafariandev:
> I am wondering if caddy is actually needed here for Go-blueprint? The existing frameworks/options creates a basic RESTFUL API with a http server. I dont think Caddy is appropriate here?

Caddy is a full-featured HTTP server like Apache/nginx.
You can also create plugins to caddy to add additional features to the base feature set.
My company uses golang and caddy. I thought this would be a nice fit since it is about HTTP Frameworks.
Also, your tool could be useful in our team as a starting place for a general-purpose code generation tool.
However, since caddy is a bit heavier than most of the frameworks here, I understand not wanting to include it.
I will be taking this idea and incorporating it into our own tooling, though. Thanks for this project.

PR #91, opened 2023-11-12
PR: Adding Cobra templates for CLI app generation

LukeTarr:


rustafariandev:
Could I use these templates in my rewrite?

LukeTarr:
@rustafariandev yeah go for it! You mean the make file and cobra root and stuff? Sure thing. It might not be the best way to showcase cobra since I made these after learning cobra this weekend haha. Glad they’re of use to you!

cmd/program/program.go
```diff
@@ -189,25 +176,79 @@ func (p *Project) CreateMainFile() error {
...
-		log.Printf("Error injecting server.go file: %v", err)
-		cobra.CheckErr(err)
-		return err
-	}
+		// inject makefile template
+		makeFileTemplate := template.Must(template.New("makefile").Parse(string(tpl.MakeTemplate())))
+		err = makeFileTemplate.Execute(makeFile, p)
+		if err != nil {
+			return err
+		}
```
Melkeydev, line 188: format?

cmd/program/program.go
```diff
@@ -100,6 +102,11 @@ func (p *Project) createFrameworkMap() {
 		packageName: echoPackage,
 		templater:   tpl.EchoTemplates{},
 	}
+
+	p.FrameworkMap["cobra"] = Framework{
```
Melkeydev, line 106: This is in the same prompt right now for the WEB frameworks 
Adding Cobra is entirely different - it spins up a completely new structure for an app

Melkeydev (commented):
Overall, this is a good PR however 
i dont think this is the direction I want to go with blueprint right now
The issue is if we add a CLI app, then that opens the flood gates for a bunch of different types of apps.
This sort of thing can get messy and as classic:
"boil the ocean"

LukeTarr:
Yeah no worries, I can go ahead and close out the PR if you don't want blueprint to go in that direction. If you do in the future, feel free to reuse any of the templates and stuff.

PR #94, opened 2023-11-13
Non-interactive environment breaks matrix workflows

CamPen21:
# Error

Matrix breaks with `error: open /dev/tty: no such device or address`

Seems like under the hood Matrix runs a non-interactive shell. Because the CLI is being run to generate the artifacts to be linted it will error out.

# Fix

Using`script` binary to fake a tty so that it doesn't complain.

### DIFF
--- a/.github/workflows/generate-linter.yml
+++ b/.github/workflows/generate-linter.yml
@@ -31,7 +31,8 @@ jobs:
         with:
           go-version: ${{ matrix.goVersion }}
       - name: build templates
-        run: go run main.go create -n ${{ matrix.framework }} -f ${{ matrix.framework}}
+        run: |
+          script -q /dev/null -c "go run main.go create -n ${{ matrix.framework }} -f ${{ matrix.framework}}" /dev/null
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-linter.yml
```diff
@@ -26,12 +26,9 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - name: Setup Go
-        uses: actions/setup-go@v4
-        with:
-          go-version: ${{ matrix.goVersion }}
       - name: build templates
-        run: go run main.go create -n ${{ matrix.framework }} -f ${{ matrix.framework}}
+        run: |
+          script -q /dev/null -c "go run main.go create -n ${{ matrix.framework }} -f ${{ matrix.framework}}" /dev/null
```
Melkeydev, line 31: Hm - this does look a bit hacky?
@mimatache i will leave you to look at this and give some suggestions

mimatache: @Melkeydev This is a good workaround for entering interactive mode with the application, though it is a hack around a GitHub limitation. 
Another option would be to simply not enter interactive mode for non interactive calls to the program. I have opened #96 for this. 
You can choose which you would prefer.

Melkeydev:
Since there are two PR's addressing the same issue - i am going to go with #96

PR #96, opened 2023-11-13
feat: add spinner v4

mimatache:
The previous implementation for the spinner was entering interactive mode all the time, this caused GHA that would run `go-blueprint` at any point  to fail. 
This affected the generate projects linting as of now, but any action that would use the `go-blueprint create` would have had the same issue.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -16,9 +16,6 @@ import (
 	"github.com/melkeydev/go-blueprint/cmd/ui/textinput"
 	"github.com/melkeydev/go-blueprint/cmd/utils"
 	"github.com/spf13/cobra"
-	"log"
-	"os"
-	"strings"
 )
 
 const logo = `
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -183,6 +183,8 @@ var createCmd = &cobra.Command{
 			nonInteractiveCommand := utils.NonInteractiveCommand(cmd.Use, cmd.Flags())
 			fmt.Println(tipMsgStyle.Render("Tip: Repeat the equivalent Blueprint with the following non-interactive command:"))
 			fmt.Println(tipMsgStyle.Italic(false).Render(fmt.Sprintf("• %s\n", nonInteractiveCommand)))
+			project.Exit = true
+			project.ExitCLI(tprogram)
 		}
 	},
 }
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -7,6 +7,7 @@ import (
 	"github.com/melkeydev/go-blueprint/cmd/program"
 	"github.com/melkeydev/go-blueprint/cmd/steps"
 	"github.com/melkeydev/go-blueprint/cmd/ui/multiInput"
+	"github.com/melkeydev/go-blueprint/cmd/ui/spinner"
 	"github.com/melkeydev/go-blueprint/cmd/ui/textinput"
 	"github.com/melkeydev/go-blueprint/cmd/utils"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


Melkeydev:
image

There is an issue going on with the go-routines.
Potentially not being closed but once the true boolean flag gets passed into the channel, this message should no longer appear after the project was generated

cmd/create.go
```diff
@@ -128,21 +129,37 @@ var createCmd = &cobra.Command{
 		}
 
 		project.AbsolutePath = currentWorkingDir
-
-		// This calls the templates
+		spinStatus := make(chan struct{})
```
Melkeydev, line 132: I think the reason is we dont actually pass this channel.
This is not properly listening to the status of the spinner, so on line 149 we are listening but are not getting any value
indicating the the spinner status completed.

cmd/create.go
```diff
@@ -128,21 +129,37 @@ var createCmd = &cobra.Command{
...
+		go func() {
+			if isInteractive {
+				tprogram = tea.NewProgram(spinner.InitialModelNew())
+					if _, err := tprogram.Run(); err != nil {
+						cobra.CheckErr(err)
+					}
+			}
+		}()
 		err = project.CreateMainFile()
+		close(spinStatus)
```
Melkeydev, line 143: Also, we close the channel here but listen to it on line 149

Melkeydev (changes requested):
The channel right now is not properly configured

CamPen21:
@mimatache Can we also get rid of lines 29 to 32 from the file [.github/workflows/generate-linter.yml.](url)
It does the same thing for every matrix arm and it generates errors.

cmd/create.go
mimatache: I'm not sure why I left this here, it was here in the original implementation and kind of left it. removed it now.

mimatache: Nevermind, I figured out what is wrong.

cmd/create.go
```diff
@@ -128,9 +130,26 @@ var createCmd = &cobra.Command{
...
+				}
+			}
+		}()
 		err = project.CreateMainFile()
+		// once the create is done, stop the spinner
+		if isInteractive {
+			spinner.Quit()
+		}
+		// wait for the spinner to finish
+		wg.Wait()
```
MohammadAlhallaq, line 152: Actually, that's an interesting approach. But I want to get something clear here; correct me if I'm wrong. First, we initialize the spinner model, then add one to the wait group indicating that we have one go-rotune we need to wait for. After that, we start the spinner in a goroutine, and then we begin creating the main file. Now, the point is, does the goroutine actually get marked as done after using spinner.Quit()? and after that wg.Wait() unblock the main routine? Or am I missing something here? Because I didn't totally understand how the goroutine is going to get marked as done after we sta

mimatache: I'm not sure what you mean by marked as done.

Here is what is going to happen (assuming this is in interactive mode):

- We initiate a wait group and add 1 to the wait for the spinner
- We start the spinner - the spinner function is a blocking function that does not actually finish running until you call `Quit()` on it.
- We then call `Quit()` on it, which ends the run previously started in the above go routine.
- Once that function exits, the defer is called for the `wg.Done()`
- Since we were only waiting for 1 item to finish, `wg.Wait()` considers everything is done (since the go routine existed) and the program resumes it's normal execution

mimatache:
> The channel right now is not properly configured

Removed the channel as it was not needed.

Melkeydev:
I am wondering if it makes sense to also include a "Finished" when the job is done?
Right now when the job completes the "preparing.." just stays there on the CLI.

Also:
Nit - the spacing is a bit too big?

image

Melkeydev (approved):
LGTM

PR #97, opened 2023-11-13
fix: string validation for flag -name

### LINKED ISSUE #86: Passing in go-blueprint create -n -f chi creates a dir named -f
image

image

We need to fix this lul

stgonzales:
fix #86


Melkeydev:
Another PR was merrged that handels this case!

PR #99, opened 2023-11-13
Feature/Tests to Generated Projects

20pa5a1210:
added tests to generated projects. just used standard go testing library because using testify or any other 3rd party libraries
need to install that package we don't know user want a testing library or not. i mean most people usually don't write tests tho. so sticking with standard testing good instead of using a 3rd party one

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -49,6 +49,7 @@ type Templater interface {
 	Routes() []byte
 	RoutesWithDB() []byte
 	ServerWithDB() []byte
+    TestHandler() []byte
 }
 
 type DBDriverTemplater interface {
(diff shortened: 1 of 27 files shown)


cmd/program/program.go
```diff
@@ -37,6 +37,7 @@ type Templater interface {
 	Main() []byte
 	Server() []byte
 	Routes() []byte
+    TestHandler() []byte
```
Melkeydev, line 40: nit: format?

cmd/program/program.go
```diff
@@ -49,6 +50,7 @@ var (
 
 	cmdApiPath         = "cmd/api"
 	internalServerPath = "internal/server"
+    testHandlerPath    = "tests"
```
Melkeydev, line 53: nit: format

Melkeydev (approved):
This looks dope - i have a few nits that I will leave up to you to address.
And OMG tests??!?!?!?!!?!?

20pa5a1210:
Fixed Format nits

PR #100, opened 2023-11-13
fix 🐛: completion in homebrew

muandane:
added scripts dir that generated temp shell completions to be used later on for homebrew potentially


scripts/completions.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -e
+rm -rf completions
+mkdir completions
+for sh in bash zsh fish; do
+  go run main.go completion "$sh" >"completions/goreleaser.$sh"
```
MitchellBerend, line 6: Is this ran on the users computer? 

I'm not sure how homebrew works but if it does get ran on their machine it requires the go compiler which might not be available.

muandane: Check .goreleaser.yml before hooks

Melkeydev (approved):
LGTB

PR #108, opened 2023-11-15
Feature/GitHubAction Templates for generated projects

Ujstor:

## Feature

I added a step for users to choose workflow templates for Go code tests when there is a push or pull request (PR) and 'goreliser' template used in this repository. I like having this feature in my projects. If you find it useful, I'm glad that I can contribute to this fire repo

## Description of Changes: 

- Added an additional step after selecting the framework
- Injected release.yml , go-test.yml & .goreleaser.yml in root
- Modified the necessary code to support this feature

## Checklist

- [ ] I have self-reviewed the changes being requested

blueprint_1


Melkeydev (changes requested):
I think this is a really cool feature - however:

- I am super hesitant in adding a third step just for work flows. It seems almost out of scope for what I had initially wanted blueprint to be.
- This feature gave me an idea, maybe we can release a blueprint lite and blueprint full version that can include multiple steps into the process?
- There are merge conflicts, but I am not sure if i necessarily want this feature inside of blueprint right now. Sorry.

20pa5a1210:
instead of adding a whole new step why don't we keep it flag only if user wants workflows they just use a flag  `--workflows` something like that.

Ujstor:
I'm glad that the feature interests you, and we can implement it if you decide to do so. However, I find adding it without user consent to be tricky. When users open a new project that has just been created, the large number of files can be confusing. Lite and full versions  or @20pa5a1210's idea(I like it); if needed, we can find a solution.

Melkeydev:
I was actually think about this recently and there was an idea:
we can add a --advanced flag that will give a lot more options for users to tinker/create with.
That way the lite version will continue to be the way it is now, and the advanced flag can have a ton of new features like this one. 
Thoughts?

Ujstor:
I like the idea. I saw bsquidwrd propose this on Discord. What should I do with this PR? I can resolve the conflict and keep it up to date. However, I'm not sure that I can implement this feature alone. Some help would be nice, or I need more time to understand the logic behind it.

PR #109, opened 2023-11-16
Fix/air: Deprecated GOPATH reliability

### LINKED ISSUE #74: Current `make watch` target fails to find `air` even after successful installation
## Description
When running `make watch` after successfully installing `air`, the Makefile still prompts "air is not installed. Do you want to install it now? (y/n)". This happens even though `air` has been installed and is available in the PATH. 

It seems like the GOPATH has been deprecated in Go 1.16 and should not be relied on.

## Environment
- Go version: 1.21.0
- OS: Ubuntu 22.04

## Additional Information
Here is the `watch` target in the Makefile:

```makefile
watch:
	@if [ -x "$(GOPATH)/bin/air" ]; then \
	    "$(GOPATH)/bin/air"; \
		@echo "Watching...";\
	else \
	    read -p "air is not installed. Do you want to install it now? (y/n) " choice; \
	    if [ "$$choice" = "y" ]; then \
			go install github.com/cosmtrek/air@latest; \
	        "$(GOPATH)/bin/air"; \
				@echo "Watching...";\
	    else \
	        echo "You chose not to install air. Exiting..."; \
	        exit 1; \
	    fi; \
	fi

### LINKED ISSUE #111: launching air through makefile not working
## Abstract

When using the `make watch` command I get the response `air is not installed. Do you want to install it now? (y/n)` Even though air is installed and is in my `$GOPATH/bin`.

## Issue Replication

Machine type: darwin amd64 (206 macbook pro)

## Expected Behavior

With my environment laid out as it is this command should run air the way it does when I use the `air` command

andrerocco:

## Problem/Feature

Closes #74;
Closes #111;

## Description of Changes: 

- Changed `make watch` to rely on command null command respons instead to check if `air` is installed of checking for GOPATH.

## Checklist


narasaka:
this should close #111 too 👍

andrerocco:
> this should close #111 too 👍

It probably does. I will close this one and open a new PR with to clean this one.

PR #113, opened 2023-11-17
Feature/Docker Compose for a quick database spin-up

Ujstor:

## Feature

Hi, I've added Docker Compose for the database when the user chooses one of the four drives.
There is no way to spin up SQLite with Docker. I created a simple app in Go that creates 'test.db' in the root of the project. For now, this is in my DockerHub. For security reasons, you can't push it to your repository. Here is code:

```go
package main

import (
	"database/sql"
	"fmt"
	"log"
	"net/http"

	_ "github.com/mattn/go-sqlite3"
)

func main() {
	db, err := sql.Open("sqlite3", "/app/db/test.db")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	_, err = db.Exec(`CREATE TABLE IF NOT EXISTS test_data (id INTEGER PRIMARY KEY, name TEXT);
	                  INSERT INTO test_data (name) VALUES ('Sample Data 1'), ('Sample Data 2');`)
	if err != nil {
		log.Fatal(err)
	}

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		rows, err := db.Query("SELECT id, name FROM test_data")
		if err != nil {
			http.Error(w, err.Error(), http.StatusInternalServerError)
			return
		}
		defer rows.Close()

		for rows.Next() {
			var id int
			var name string
			err = rows.Scan(&id, &name)
			if err != nil {
				http.Error(w, err.Error(), http.StatusInternalServerError)
				return
			}
			fmt.Fprintf(w, "ID: %d, Name: %s\n", id, name)
		}
	})

	log.Println("Server started on port 6969")
	log.Fatal(http.ListenAndServe(":6969", nil))
}
```

```bash
FROM golang:1.21-alpine

WORKDIR /app

RUN apk add --no-cache gcc musl-dev

ENV CGO_ENABLED=1

COPY . .

RUN go mod init sqlite-test && \
    go get github.com/mattn/go-sqlite3 && \
    go build -o main .

EXPOSE 6969

CMD ["./main"]
```
## Description of Changes: 

- Docker compose DB templates

## Checklist


Melkeydev:
```There is no way to spin up SQLite with Docker. ```

I wonder if this can be explored - I will not approve a PR that is pulling test Docker hubs, that is just a smell.

A workaround is to have a Dockerfile for sqlite, and a docker-compose that runs it. Something like this:

```
FROM alpine:latest

RUN apk add --no-cache sqlite

RUN mkdir /db
WORKDIR /db

CMD ["sqlite3"]
```

and in a docker-compose file:
```
version: '3'

services:
  sqlite:
    build: .
    volumes:
      - ./data:/db
    stdin_open: true
    tty: true
    ```

cmd/template/dbdriver/files/docker-compose/sqlite.tmpl
```diff
@@ -0,0 +1,9 @@
+version: '3.8'
+
+services:
+  sqlite:
+    image: ujstor/sqlite-test
```
Melkeydev, line 5: This is a bit of a smell.
I mentioned in an earlier comment it might be better to just have a dedicated Docker file for this

cmd/template/dbdriver/files/docker-compose/sqlite.tmpl
```diff
@@ -0,0 +1,9 @@
...
+    ports:
+      - "6969:6969"
```
Melkeydev, line 7: meme?

Melkeydev (changes requested):
Mostly on the SQLite implementation

cmd/template/dbdriver/files/docker-compose/sqlite.tmpl
Ujstor: I am wondering if there's a need for SQLite. If you curl localhost:8080/health and there is no error, and if the DB_URL env is set (for example, to ./test.db), a curl command would create db in the root directory

Ujstor:
Let me know if you agree with this comment above.
I made a small hack to ensure everything works as expected. I left an empty SQLite template to maintain the simplicity of the implementation. The empty docker-compose file is deleted at the end if the user chooses SQLite driver. I hope this is okay, if not, I will find another solution.

Ujstor:
You can ignore the most recent comment. I rewrote the logic and removed the SQLite template entirely.

cmd/program/program.go
```diff
@@ -211,10 +240,25 @@ func (p *Project) CreateMainFile() error {
...
-			log.Printf("Error injecting server.go file: %v", err)
+			log.Printf("Error injecting database.go file: %v", err)
 			cobra.CheckErr(err)
 			return err
 		}
+
+		
+	}
+
+	// cretae correct docker compose for the selected driver
```
Melkeydev, line 251: typo

cmd/program/program.go
```diff
@@ -211,10 +240,25 @@ func (p *Project) CreateMainFile() error {
...
+	if p.DBDriver != "none" && p.DBDriver != "sqlite" {
+    	p.createDockerMap()
+    	p.Docker = p.DBDriver
+
+    	err = p.CreateFileWithInjection(root, projectPath, "docker-compose.yml", "db-docker")
+    	if err != nil {
+    	    log.Printf("Error injecting docker-compose.yml file: %v", err)
+    	    cobra.CheckErr(err)
+    	    return err
+    	}
```
Melkeydev, line 261: Format tab to the right

cmd/program/program.go
```diff
@@ -211,10 +240,25 @@ func (p *Project) CreateMainFile() error {
...
+			log.Printf("Error injecting database.go file: %v", err)
 			cobra.CheckErr(err)
 			return err
 		}
+
+		
+	}
+
+	// cretae correct docker compose for the selected driver
+	if p.DBDriver != "none" && p.DBDriver != "sqlite" {
```
Melkeydev, line 252: We might want to handle the case where they choose sqlite and we reply saying we cannot create a docker-compose instance, or something like that

Melkeydev (changes requested):

Melkeydev:
Can we also include spinning up the docker-compose instance in a make command?

Ujstor:
I made the requested changes. Now, if a user chooses SQLite, they will receive this message. I messed up the formatting, which is the reason for so many commits.

mintty_3utHZQC4US

cmd/utils/utils.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/spf13/pflag"
 )
 
-const ProgramName = "go-blueprint"
+const ProgramName = "go-blueprint create"
```
20pa5a1210, line 13: i think you remove create from here. this non-interactive command tip message issue resolved in #123

20pa5a1210 (changes requested):

cmd/utils/utils.go
Ujstor: tnx

cmd/template/framework/files/makefile.tmpl
```diff
@@ -11,6 +11,14 @@ build:
 run:
 	@go run cmd/api/main.go
 
+# Create DB container
+docker-run:
+	@docker compose up -d
```
Melkeydev: this seems to be a plugin version for docker.
Lets go with ol' reliable:
`docker-compose up`

cmd/template/framework/files/makefile.tmpl
```diff
@@ -11,6 +11,14 @@ build:
...
+
+# Showdown DB container
+docker-down:
+	@docker compose down
```
Melkeydev: Same thing here:
`docker-compose down`

cmd/template/framework/files/makefile.tmpl
Melkeydev: I am learning now that `docker-compose` is V1 and was deprecated in June 2023.
This is a problem because I dont think it is backwards compatible from V2..

I think right now it is safer to keep it in V1 though... IDK

cmd/template/framework/files/makefile.tmpl
```diff
@@ -11,6 +11,14 @@ build:
 run:
 	@go run cmd/api/main.go
 
+# Create DB container
+docker-run:
+	@docker-compose up
+
```
Melkeydev, line 17: Thinking about this i believe the best way to handle this is the following:

docker-run:
        # this uses V2
	@docker compose up
        if fail:
            echo "this requires Docker V2 - use docker-compose up"
        else:
            do nothing continue with execution
            
 
I think this is the best path since this is going to use a non deprecated dependency (v1 docker-compose).

cmd/template/docker/files/docker-compose/mongo.tmpl
```diff
@@ -0,0 +1,15 @@
+version: '3.8'
+
+services:
+  mongo:
+    image: mongo:latest
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: ${DB_USERNAME}
+      MONGO_INITDB_ROOT_PASSWORD: ${DB_ROOT_PASSWORD}
+    ports:
+      - "${DB_PORT}:27017"
```
Melkeydev, line 10: *Adding a general comment here that applies to all the DB templates*
The thing I am noticing is that if the user does not have their .env set up - obviously these will not work.

So what I am thinking is to do what create-t3-app does, which is basically spin up a .env.example file that has placeholders or somthing that then a user can use to remake into their own .env

That way they can set it up themselves and not have to worry about it not working.
What do you think?
Mayve in the .env.example we can also pre-fill these values to show a user what they will need.

Melkeydev (changes requested):

Ujstor:
Requested changes are ready.

Ujstor:
@Melkeydev 
I found a bug, or rather something that's not exactly a bug per se. If a user creates a Postgres db with Docker Compose for the first time and binds the environment variables, everything works perfectly. A volume is automatically created for data persistence. However, when these variables are changed, Postgres will not initialize another database, user, or password.

This is the message:
'PostgreSQL Database directory appears to contain a database; Skipping initialization.'

Performing a curl /health will break the app, but only if the environment variables are changed on an existing volume. The user needs to delete this volume, or change the volume name in docker-compose.

PostgreSQL doesn't automatically reinitialize the database when it finds existing data in the volume. This behavior is different from other databases. MySQL or MongoDB are more flexible in handling changes in environment variables.

cmd/template/framework/files/makefile.tmpl
```diff
@@ -11,6 +11,24 @@ build:
...
+# Create DB container
+docker-run:
+	@if docker compose up 2>/dev/null; then \
+		: ; \
+	else \
+		echo "Falling back to Docker Compose V1"; \
+		docker-compose up; \
+	fi
+
+# Showdown DB container
```
Melkeydev: typo

cmd/template/framework/files/README.md.tmpl
```diff
@@ -23,6 +23,16 @@ run the application
 make run
 ```
 
+Create DB container
+```bash
+make docker-run
+```
+
+Showdown DB container
```
Melkeydev: typo

Melkeydev (approved):
This.
Looks.
Mother
Fucking.
DELICIOUS.
To Me.

PR #120, opened 2023-11-19
fix: change cmd in .air.toml to build instead of run

narasaka:

## Problem/Feature

Prior to this commit, running `make watch` will result in the program running in the background without any way of killing it other than looking for its PID and killing that. This changes the "cmd" in the ".air.toml" file to run `go build -o ./tmp/main ./cmd/api` instead of `make run`.

Running `make watch` prior to this commit:
image

It is stuck on "building..." but the program is already running in the background. ctrl+c kills air but does not kill the program running.

Running `make watch` with this commit:
image

program is now running as expected and doing ctrl+c kills the program properly.

## Description of Changes: 

- Change `cmd` in `.air.toml` to ~`make build`~ `go build -o ./tmp/main ./cmd/api`
- ~Change `make build`'s output to `tmp/main` instead of `main`~

## Checklist

### DIFF
--- a/cmd/template/framework/files/air.toml.tmpl
+++ b/cmd/template/framework/files/air.toml.tmpl
@@ -5,7 +5,7 @@ tmp_dir = "tmp"
 [build]
   args_bin = []
   bin = "./tmp/main"
-  cmd = "make run"
+  cmd = "make build"
   delay = 1000
   exclude_dir = ["assets", "tmp", "vendor", "testdata"]
   exclude_file = []
--- a/cmd/template/framework/files/air.toml.tmpl
+++ b/cmd/template/framework/files/air.toml.tmpl
@@ -5,7 +5,7 @@ tmp_dir = "tmp"
 [build]
   args_bin = []
   bin = "./tmp/main"
-  cmd = "make build"
+  cmd = "go build -o ./tmp/main ./cmd/api"
   delay = 1000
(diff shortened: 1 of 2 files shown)


cmd/template/framework/files/air.toml.tmpl
```diff
@@ -5,7 +5,7 @@ tmp_dir = "tmp"
 [build]
   args_bin = []
   bin = "./tmp/main"
-  cmd = "make run"
+  cmd = "make build"
```
20pa5a1210, line 8: `cmd = "go build -o ./tmp/main ./cmd/api"`

we can run directly like this instead using build command. imo making build file in tmp folder not a good practice

cmd/template/framework/files/makefile.tmpl
```diff
@@ -5,7 +5,7 @@ all: build
 
 build:
 	@echo "Building..."
-	@go build -o main cmd/api/main.go
+	@go build -o tmp/main cmd/api/main.go
```
20pa5a1210, line 8: i think mostly people expect  build file to be in root of the folder so we should keep it that way

20pa5a1210 (changes requested):

20pa5a1210 (approved):

Melkeydev (approved):
lgtm

Issue #124, opened 2023-11-19
go-blueprint version string is empty on output instead of it printing the version


brotherthegrest:
when go-blueprint version is written the output is just  `Go Blueprint CLI version`
instead of printing the proper output `Go Blueprint CLI version 0.3.0`

FMinister:
I would say that this is an error in the [.goreleaser.yml](https://github.com/Melkeydev/go-blueprint/blob/main/.goreleaser.yml) file. There, `github.com/melkeydev/go-blueprint/cmd.GoBlueprintVersion={{.Version}}` should be replaced with `cmd.GoBlueprintVersion={{.Version}}`.

limesten:
I believe the problem is that when you install with `go install github.com/melkeydev/go-blueprint@latest`

It will build from the source code rather than using the pre-built binary, and therefore the version variable won't be populated with the ldflag

Working on a fix for this one

PR #135, opened 2023-11-22
fix: version command

limesten:

## Problem/Feature

- fixes issue: #124 

the command `go-blueprint version` works fine when downloading the pre-build binary from  a release.
However, if doing `go install github.com/melkeydev/go-blueprint@latest` it seems to result in an empty `GoBlueprintVersion` string.

## Description of Changes: 

lost a few hairs on my head investigating this but:
- Added a function called `getGoBlueprintVersion` 
If the version var has been set by goreleaser it will use that
else we will use `ReadBuildInfo` from the runtime/debug package (which gets the version from git tags)
also added a fallback which just prints version unknown

## Checklist


Melkeydev (commented):
I am trying out this branch and I am getting the following:
```
  ~/Documents/Personal/Go/test                                                                                                                                                             
❯ go-blueprint version
Go Blueprint CLI version (devel)
```

Is this expected? I am unsure what i did

cmd/version.go
```diff
@@ -5,20 +5,32 @@ package cmd
...
 
 	"github.com/spf13/cobra"
 )
 
 // GoBlueprintVersion is the version of the cli to be overwritten by goreleaser in the CI run with the version of the release in github
 var GoBlueprintVersion string
 
+func getGoBlueprintVersion() string {
+	if len(GoBlueprintVersion) != 0 {
+		return GoBlueprintVersion
```
Melkeydev, line 18: We can probably define this inline and not have a separate variable declared on line 14

cmd/version.go
```diff
@@ -5,20 +5,32 @@ package cmd
...
+	}
+	if info, ok := debug.ReadBuildInfo(); ok {
+		return info.Main.Version
+	}
+	return "unknown"
```
Melkeydev, line 23: Nit: Maybe we can return something else rather than "unknown" since that is a big ambiguous

Melkeydev (changes requested):

limesten:
> I am trying out this branch and I am getting the following:
> 
> ```
>   ~/Documents/Personal/Go/test                                                                                                                                                             
> ❯ go-blueprint version
> Go Blueprint CLI version (devel)
> ```
> 
> Is this expected? I am unsure what i did

For the info.Main.Version to embed the version from the git tag it needs to be installed like:
`go install github.com/melkeydev/go-blueprint@latest` (or with the specific version instead of latest)

If we install like that and then run `go version -m go-blueprint` on the binary in the bin folder we get:

image
<br><br>
However, if we run `go install` locally with the source code we will get `(devel)`
<br><br>

image

cmd/version.go
limesten: But GoBlueprintVersion needs to be on package level to be able to be set by the ldflag in .goreleaser.yml?
Or am I misunderstanding you?

cmd/version.go
limesten: true, will try to come up with something better

muandane:
i'm a bit confused to why it's not working also at build i tested the build locally 
image

limesten:
Thoughts on these changes?

- Added more comments 

- Added an additional fallback using git revision hash and build time

- Also changed to a less ambiguous output if no version info is available

Also for additional info about getting (devel) from local builds with debug.ReadBuildInfo() there are these related issues on the golang repo:
https://github.com/golang/go/issues/29228
https://github.com/golang/go/issues/50603

Melkeydev:
@limesten I am still investigating this but it seems it still does not work
image

limesten:
> @limesten I am still investigating this but it seems it still does not work image

I assume you are checking out the pr and then running go install. I've not found a good way to include the version info in this case. 

It will work if:

1) you make a release with goreleaser (for example goreleaser `build --single-target --skip-validate --clean` as a local test release)
(or manually enter a ldflag when running go build as @muandane did)

2) you install like this:  `go install github.com/melkeydev/go-blueprint@latest` or @v.0.4.1 etc
In this case the part after the @ is what gets added to the buildinfo so it will be accessible from the version command

The weird thing with 2) is that u cant't rly test it locally beforehand (as far as I can see at least)

Ujstor:
@Melkeydev This is working. I created a test repo, pushed the tag, the build passed and the version was correct. LGTM
Screenshot from 2023-12-19 07-22-20

Melkeydev (approved):
LGTM

PR #136, opened 2023-11-22
Use filepath

rustafariandev:

## Problem/Feature

The use of Sprintf for joining file paths makes it impossible to run on windows.

## Description of Changes: 

- use filepath.Join instead of Sprintf to improve portability

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+go-blueprint
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -8,6 +8,7 @@ import (
 	"html/template"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -59,7 +60,7 @@ type Templater interface {
 	Routes() []byte
 	RoutesWithDB() []byte
 	ServerWithDB() []byte
-    	TestHandler() []byte
+	TestHandler() []byte
 }
 
 type DBDriverTemplater interface {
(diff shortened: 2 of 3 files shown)


Melkeydev:
Wait, maybe i am just too ooga booga brain but why does using  Sprintf for joining file paths make it impossible to work with on Windows?

.gitignore
```diff
@@ -0,0 +1 @@
+go-blueprint
```
Melkeydev, line 1: Why are we ignoring this? Is this the binary after `go build` ?

Melkeydev (approved):
LGTM
 I tested this out on my local instance , - looks good.
 However I have some comments that are non-blocking

.gitignore
rustafariandev: To avoid seeing it when doing a 
git status

Also, not to mistakenly add it to the repo on a commit.

It is not important. Would you want me to remove it, or should we keep it?

rustafariandev:
> Wait, maybe i am just too ooga booga brain but why does using  Sprintf for joining file paths make it impossible to work with on Windows?

https://www.minitool.com/news/backslash-vs-forward-slash.html

.gitignore
rustafariandev: yes it is the binary after go build

Melkeydev (changes requested):
There is some linting changes failing - we need to check these out!

rustafariandev:
Done!

Melkeydev (approved):
LGTM

Issue #140, opened 2023-11-25
[Feature Request]  Remove .env and add .env to .gitignore


Melkeydev:
### Tell us about your feature request

Since we are adding .env.example we no longer need a .env file and additionally we should include this as one of the files to ignore

### Disclaimer

- [X] I agree

Ujstor:
@Melkeydev 

I don't think it's a good idea to remove the .env file completely. It is a two-part file, created from globalenv.tmpl and a specific DB template. Maybe we can remove the DB-specific variables. If we keep the current configuration, that's good, or alternatively, we could remove the .env.example and bind default variables in .env so that Docker Compose works out of the box.

What do you think is the best approach?

```
#global 
PORT=8080
APP_ENV=local 

#DB specific
DB_HOST=
DB_PORT=
DB_DATABASE=
DB_USERNAME=
DB_PASSWORD=
DB_ROOT_PASSWORD=
```

PR #145, opened 2023-11-28
Fix/env

SudoSurya:

## Problem/Feature

- removed example env files and binded env values to default env file directly so user can spin docker direclty without changing env files #140 
- separate makefile for none db projects and sqlite(sqlite doesnt have docker file adding docker commands in make file of sqlite project doesnt make sense )

## Checklist


Ujstor:
I tested PR and it's working perfectly. I am in favor of this solution, binding environment variables so that everything works out of the box is a good approach. I am interested what Melkey thinks about this one.

Melkeydev:
@SudoSurya Some of the tests are breaking here for deployment!

PR #151, opened 2023-12-12
feat: env fix

SudoSurya:

## Problem/Feature

- removed example env files and binded env values to default env file directly so user can spin docker direclty without changing env files #140 
- separate makefile for none db projects and sqlite(sqlite doesnt have docker file adding docker commands in make file of sqlite project doesnt make sense )

## Checklist


SudoSurya:
i closed #145 and opened this pr (no new changes )

cmd/program/program.go
```diff
@@ -317,11 +309,20 @@ func (p *Project) CreateMainFile() error {
...
-		return err
+	if p.DBDriver == "sqlite" || p.DBDriver == "none" {
+		// inject makefile template
+		makeFileTemplate := template.Must(template.New("makefile").Parse(string(framework.NonDbMakeFileTemplate())))
+		err = makeFileTemplate.Execute(makeFile, p)
+		if err != nil {
+			return err
+		}
+	} else {
+		// inject makefile template
```
Melkeydev, line 320: nit: Maybe change this to say this one includes the commands for Databases as well

Melkeydev (approved):
One small nit

Melkeydev:
One thing I noticed is that if a user choose SQLite, they are kind of left thinking what should they next, since nothing is generated.

Melkeydev (approved):
small nit but up to you - LGTM

SudoSurya:
> One thing I noticed is that if a user choose SQLite, they are kind of left thinking what should they next, since nothing is generated.

makefile will be generated for sqlite also but it doesn't not include any commands related to database (we have only docker commands in other databases since we don't have any docker commands for sqlite we generating normal makefile with doesnt include docker commands)

PR #154, opened 2023-12-16
[Enhancement]: Use fiber.Map instead of map object

NimishKashyap:

## Problem/Feature

GoFiber package recommends using fiber.Map{} type which is an alias of map interface for easier JSON handling.

## Description of Changes: 

- Changed `map[string]string` type for `HelloWorldHandler` response to standard `fiber.Map{}` as recommended by the package.

## Checklist

- [X] I have updated the documentation (if applicable)


Melkeydev (approved):
LGTM

PR #156, opened 2023-12-17
Add Microsoft Sql Server driver

AndreaGero:
I've recently needed to integrate SQL Server driver connection on my Project, so I decided to add a new Driver choice.


Melkeydev:
This is a bit of an opinionated change, adding another database driver.
I am not entirely against it, but, I would perhaps change out one of the current drivers with this one and see what others say.
One other note, we recently are moving away from a .env example and just have the values update directly the .env file

PR #164, opened 2023-12-27
Added templ generated files in gitignore

saphal1998:

## Problem/Feature

The gitignore template does not exclude the generated templ files. In my opinion, this should be excluded from VCS. 

I'm not very sure about what "best practice" dictates here, but to me it makes sense that generated files should not be in version control and should be generated everytime?

Sorry if this is an unwelcome change because it's so small, but I think I add it everytime I use it, so figured it might be worth just making a PR for it.

## Description of Changes: 

- Added the regex for excluding any files of files ending with `_templ.go`.

## Checklist

- [X] I have updated the documentation (if applicable)

### DIFF
--- a/cmd/template/framework/files/gitignore.tmpl
+++ b/cmd/template/framework/files/gitignore.tmpl
@@ -24,3 +24,6 @@ tmp/
 
 # .env file
 .env
+
+# templ generated files
+**/*_templ.go


michaelhvisser:
@saphal1998 I was thinking this also needed to be added to the gitignore. Why did you end up closing this?

saphal1998:
I am actually trying to make a small-ish project using Templ, will compile all findings and add them in one go, that was my plan

PR #172, opened 2024-01-19
fix: Adding templ specific config to air.toml

### LINKED ISSUE #171: [Bug] Using Air and templ (hot/live-reloading doesn't work)
### What is the problem?

So I went with a go-blueprint using Chi and then standard setup using templ files. When I try to execute anything with Air my Go code reruns but either it crashes my servers (it stops going to run step), doesn't reload my templ files or keeps reloading out of control.

### Operating System

MacOs

### Architecture Version (x86, x64, arm, etc)

m1

### Steps to reproduce

1. go-blueprint create --advanced
2. [x] chi
3. [x] postgres 
4. [x] HTMX/templ files 

### Relevant log output

_No response_

Waldeedle:

## Problem/Feature

Adding templ specific config to the files watched by air and also ensuring that the makefile commands are used for air.
Currently there is not a lot of complexity in the build process but as configuration increases, I would assume the team would want the makefile command to be the source of truth.
Resolves #171

## Description of Changes: 

- Added regex to ignore _templ.go files
- Added .templ files to the air toml
- Added `templ generate` to the makefiles conditionally

## Checklist

### DIFF
--- a/cmd/template/framework/files/air.toml.tmpl
+++ b/cmd/template/framework/files/air.toml.tmpl
@@ -4,17 +4,17 @@ tmp_dir = "tmp"
 
 [build]
   args_bin = []
-  bin = "./tmp/main"
-  cmd = "go build -o ./tmp/main ./cmd/api"
+  bin = "./main"
+  cmd = "make build"
   delay = 1000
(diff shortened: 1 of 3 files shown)


Melkeydev (approved):
LGTM

PR #177, opened 2024-01-25
Add redis as an option for database

schoolboybru:

## Problem/Feature

Adding Redis as an option for a database. 

## Description of Changes: 

1. Add all templates for redis
2. Make redis an option when selecting a database
3.  Update documentation 

## Checklist


cmd/template/dbdriver/files/service/redis.tmpl
```diff
@@ -0,0 +1,62 @@
...
+	err := s.db.Ping(ctx)
+
+	if err != nil {
+		log.Fatalf(fmt.Sprintf("db down: %v", err))
+	}
+
+	return map[string]string{
+		"message": "It's healthy",
+	}
+}
```
Ujstor, line 62: Ping is returning an error, and it is breaking the app when you curl the health route. 

I think this solves the issue

```go
func (s *service) Health() map[string]string {
	defer cancel()

	result, err := s.db.Ping(ctx).Result()

	if err != nil {
	}

	 if result != "PONG" {
        log.Fatalf("Unexpected ping response: %s", result)
    }

	return map[string]string{
		"message": "It's healthy",
	}
}
```

Ujstor (changes requested):
After fixing the Health func, this is ready for merge

cmd/template/dbdriver/files/service/redis.tmpl
schoolboybru: Oh good catch 👍

Ujstor (approved):
LGTM

Issue #183, opened 2024-02-12
[Feature Request] Add support for advanced features to be chosen inside flags


briancbarrow:
### Tell us about your feature request

Currently a user can only select whether or not they want to choose from the advanced features when going through the prompts. There is no way to specify which features to add from the command line flags. Adding this will allow for faster setup if someone already knows what features they want. 

This will also allow for advanced features to be included in the UI from issue #we

### Disclaimer

- [X] I agree

Marcellofabrizio:
@briancbarrow, could it be something like this?

`go-blueprint create --advanced --features htmx go-workflow`

briancbarrow:
> @briancbarrow, could it be something like this?
> 
> `go-blueprint create --advanced --features htmx go-workflow`

Yeah, that's basically what I was thinking

Marcellofabrizio:
Ok, this seems like a good first issue.

briancbarrow:
@Marcellofabrizio do you want me to assign it to you?

Marcellofabrizio:
@briancbarrow Yes, assign it to me. I had a look at the code yesterday. I'm fairly new to Go, but I'll make sure it's a nice solution 🙂

briancbarrow:
@Marcellofabrizio 👍 Let us know if you need anything.

Marcellofabrizio:
@briancbarrow I created a PR for this #184

Marcellofabrizio:
Hey @briancbarrow, I was wondering if there's something wrong in the PR since no one commented on it in a week. Did you manage to have a look?

briancbarrow:
> Hey @briancbarrow, I was wondering if there's something wrong in the PR since no one commented on it in a week. Did you manage to have a look?

I haven't had a chance to look at it yet. Sorry.

briancbarrow:
@Marcellofabrizio I finally had a chance to review it. Left some comments. Feel free to reach out if you have any questions.

PR #184, opened 2024-02-14
Adds support for advanced features to be chosen inside flags

Marcellofabrizio:

## Problem/Feature

Implementation of #183

## Description of Changes: 

- I added flags for the advanced features to be added without opening the prompt
- I replaced "AddHTMXTmpl" with "Htmx" to be more standard with the other flags. For this, I replaced all files that used the old name.
- I replaced the hardcoded calls for "AddHTMXTmpl" and "GitHubAction" with their respective new flag constants

## Checklist

PS: I'm new to Go, so I would appreciate constructive feedback 😄


cmd/create.go
```diff
@@ -170,18 +172,35 @@ var createCmd = &cobra.Command{
...
-			for key, opt := range options.Advanced.Choices {
-				project.AdvancedOptions[key] = opt
-			}
-			if err != nil {
-				log.Fatal("failed to set the htmx option", err)
+
+			featureFlags := cmd.Flag("feature").Value.String()
+			hasFeatureFlags := featureFlags != ""
+
+			if hasFeatureFlags {
```
briancbarrow, line 179:
```suggestion
			featureFlags := cmd.Flag("feature").Value.String()

			if featureFlags != "" {
```

We can remove the extra variable and just do the check right on `featureFlags`

cmd/create.go
```diff
@@ -170,18 +172,35 @@ var createCmd = &cobra.Command{
...
+				featuresFlagValues := strings.Split(featureFlags, ",")
+				fmt.Println(len(featuresFlagValues))
+				for _, key := range featuresFlagValues {
+					project.AdvancedOptions[key] = true
+				}
+
+				fmt.Println(project.AdvancedOptions)
```
briancbarrow, line 186: forgotten print statement 😃

cmd/create.go
```diff
@@ -170,18 +172,35 @@ var createCmd = &cobra.Command{
...
+				isInteractive = true
+				step := steps.Steps["advanced"]
+				tprogram = tea.NewProgram((multiSelect.InitialModelMultiSelect(step.Options, options.Advanced, step.Headers, project)))
+				if _, err := tprogram.Run(); err != nil {
+					cobra.CheckErr(textinput.CreateErrorInputModel(err).Err())
+				}
+				project.ExitCLI(tprogram)
+				for key, opt := range options.Advanced.Choices {
+					fmt.Println(key)
+					fmt.Println(opt)
```
briancbarrow, line 198: forgotten print statements

cmd/create.go
```diff
@@ -228,7 +247,7 @@ var createCmd = &cobra.Command{
 		fmt.Println(endingMsgStyle.Render("\nNext steps:"))
 		fmt.Println(endingMsgStyle.Render(fmt.Sprintf("• cd into the newly created project with: `cd %s`\n", project.ProjectName)))
 
-		if options.Advanced.Choices["AddHTMXTempl"] {
+		if options.Advanced.Choices[string(flags.Htmx)] {
```
briancbarrow, line 250: I like this change. 👍

cmd/flags/advancedFeatures.go
```diff
@@ -0,0 +1,36 @@
...
+	// Contains isn't available in 1.20 yet
+	// if AllowedProjectTypes.Contains(value) {
+	for _, advancedFeature := range AllowedAdvancedFeatures {
+		if advancedFeature == value {
+			*f = append(*f, advancedFeature)
+			return nil
+		}
+	}
+
+	return fmt.Errorf("Advanced Feature to use. Allowed values: %s", strings.Join(AllowedAdvancedFeatures, ", "))
```
briancbarrow, line 35: Go will yell at you if you have error strings that are capitalized. 🤷‍♂️ 
```suggestion
	return fmt.Errorf("advanced feature allowed values: %s", strings.Join(AllowedAdvancedFeatures, ", "))
```

Marcellofabrizio:
Hey @briancbarrow, I really appreciate the feedback, I'll be working on those comments as soon as I can. Regarding the suggested command, what should be the expected output?

briancbarrow:
> Hey @briancbarrow, I really appreciate the feedback, I'll be working on those comments as soon as I can. Regarding the suggested command, what should be the expected output?

We should either change the code to accept it like 
```
go-blueprint create --advanced true --feature htmx,githubaction
```

or like this (I think this will be easier to do).
```
go-blueprint create --advanced --feature htmx --feature githubaction
```

Marcellofabrizio:
@briancbarrow I implemented some changes that will allow us to print the non-interactive command correctly to the new format. Here's how it's looking:
image
I took the liberty to remove the `true` following the `--advanced` as it looked redundant since we can't have the flag with a false value.

I also addressed your other comments.

Marcellofabrizio:
Hey @briancbarrow, I implemented the changes to print the correct command based on the new flags. Here's how it looks: 
image

I also addressed your other comments.

cmd/create.go
Marcellofabrizio: Commited!

cmd/create.go
Marcellofabrizio: The best debbuger

cmd/create.go
```diff
@@ -170,18 +172,35 @@ var createCmd = &cobra.Command{
...
-			}
-			if err != nil {
-				log.Fatal("failed to set the htmx option", err)
+
+			featureFlags := cmd.Flag("feature").Value.String()
+			hasFeatureFlags := featureFlags != ""
+
+			if hasFeatureFlags {
+				featuresFlagValues := strings.Split(featureFlags, ",")
+				fmt.Println(len(featuresFlagValues))
```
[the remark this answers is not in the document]
Marcellofabrizio: The best debbuger

cmd/flags/advancedFeatures.go
Marcellofabrizio: Commited!

cmd/create.go
```diff
@@ -228,7 +240,7 @@ var createCmd = &cobra.Command{
 		fmt.Println(endingMsgStyle.Render("\nNext steps:"))
 		fmt.Println(endingMsgStyle.Render(fmt.Sprintf("• cd into the newly created project with: `cd %s`\n", project.ProjectName)))
 
-		if options.Advanced.Choices["AddHTMXTempl"] {
+		if options.Advanced.Choices[string(flags.Htmx)] {
```
briancbarrow, line 243: I was running through this again. I like the idea of this change, but the choice is `"Htmx"` and `string(flags.Htmx)` returns `"htmx"` so these 'next steps' never get printed. 
```suggestion
		if options.Advanced.Choices["Htmx"] {
```

briancbarrow:
> Hey @briancbarrow, I implemented the changes to print the correct command based on the new flags. Here's how it looks: image
> 
> I also addressed your other comments.

Sorry again for the late reply. Things are looking good. I did notice that now the steps for getting the htmx stuff working/generated are not printing as expected. See my comment above for that. I'm also still not getting the right output. I'm still getting 
`go-blueprint create --name my-proj --framework chi --driver mysql --advanced true` instead of the advanced features broken out. Can you confirm on your side that you are getting what is expected

Screen Shot 2024-03-05 at 9 11 05 PM

Marcellofabrizio:
- @briancbarrow Yes, I can replicate this. You're getting `go-blueprint create --name my-proj --framework chi --driver mysql --advanced true` because my code is only handling the features added in flags, not the ones selected by the steps. I didn't consider this case when I added it. I'll change it to handle both cases.

Marcellofabrizio:
Hey @briancbarrow, it was a pretty simple fix, all I had to do was apply the same logic we are doing to the other flags, which is setting the selected value to the flag after we select them.

```
for key, opt := range options.Advanced.Choices {
	project.AdvancedOptions[strings.ToLower(key)] = opt
	err := cmd.Flag("feature").Value.Set(strings.ToLower(key))
	if err != nil {
		log.Fatal("failed to set the feature flag value", err)
	}
}
```
This allows the values to be passed down to `NonInteractiveCommand` and be printed as expected.

briancbarrow (approved):
This looks good to me now.

Marcellofabrizio:
@briancbarrow Thank you!

briancbarrow:
> @briancbarrow Thank you!

Thank you for working on this!

Issue #182, opened 2024-02-08
[Feature Request] Logging support


spa5k:
### Tell us about your feature request

Something like global setup of loggers, zap/slog/zerolog etc, and optimal way of importing the logger in various places

### Disclaimer

- [X] I agree

Ujstor:
We believe that incorporating Slog or the standard logging library from Go would be the most suitable approach.

mcheviron:
> We believe that incorporating Slog or the standard logging library from Go would be the most suitable approach.

Will you add it or should someone make a pull request?
Also for whoever does it, I suggest creating an abstraction over LogAttrs instead of using the slog API. Otherwise each log will alloc on the heap, unlike Zap that uses this strategy by default

Ujstor:
This is open for anyone willing to step forward and address this enhancement. Just let me know, and I can assign it

briancbarrow:
>Also for whoever does it, I suggest creating an abstraction over LogAttrs instead of using the slog API. Otherwise each log will alloc on the heap, unlike Zap that uses this strategy by default

@mcheviron Do you have any suggested reading about this before someone takes this on?

mcheviron:
> > Also for whoever does it, I suggest creating an abstraction over LogAttrs instead of using the slog API. Otherwise each log will alloc on the heap, unlike Zap that uses this strategy by default
> 
> @mcheviron Do you have any suggested reading about this before someone takes this on?

https://pkg.go.dev/log/slog@go1.21.3#LogAttrs

The interface would be simple. A simple logger struct that exposes the same API but has a private context inside of it to enable the LogAttrs method. Then this would allow the caller to just send messages to the API for .Info, .Debug etc and the user won't feel a difference but the allocations will remain in the stack. This is how Zap does it internally. The standard slog API uses `any` which is a heap allocation. You can `go run -gcflags "-m"` and observe it


As for implementing it. How would you like implementing it? Should this logger be put in a separate file in the cmd dir or how would it work for your project? I have the code. I created it for work but i'd like to know how best is it to contribute it to your project


P.S: https://www.reddit.com/r/golang/comments/19882h2/slog_why_do_my_attrs_escape_to_the_heap/?utm_source=share&utm_medium=web2x&context=3
Not quality content but we discussed it there before

spa5k:
Rather than this, what if we use slog, then implement its handlers in various loggers, like zap etc.

```go
package main

import (
    "log/slog"

    "go.uber.org/zap"
[...]

PR #191, opened 2024-02-23
Logger

mcheviron:

## Problem/Feature

Issue #182

## Description of Changes: 

This PR introduces a logger template and updates the main template to include the logger. Here are the detailed changes:

- **Logger Template Added:** A new template for the logger has been added at:
  - `cmd/template/framework/files/main/logger.go.tmpl`
    - This template provides a basic structure for implementing logging functionality.
    - The logger is designed to be placed in the internal directory, offering an initial setup that can be extended as needed.

- **Main Template Updated:** The main template file has been updated to incorporate the logger.
  - `cmd/template/framework/files/main/main.go.tmpl`
    - Modifications include the integration of the logger template.
    - Additional comments and explanations for extending the logger have been provided.
  
- **program.go Updated:** The program.go file has been updated to inject the logger.go file in the internal path.
  - `cmd/program/program.go`

- **main.go Updated:** The main.go file has been updated to embed the logger template.
  - `cmd/template/framework/main.go`

## Checklist

- [ x ] I have self-reviewed the changes being requested
- [ x ] I have updated the documentation (if applicable)


cmd/template/framework/files/main/logger.go.tmpl
```diff
@@ -0,0 +1,89 @@
+package server
```
Ujstor, line 1: "I think a better place for the logger template is under 'template/files'

cmd/program/program.go
```diff
@@ -534,6 +534,12 @@ func (p *Project) CreateMainFile() error {
 		cobra.CheckErr(err)
 		return err
 	}
+	err = p.CreateFileWithInjection(internalServerPath, projectPath, "logger.go", "logger")
+	if err != nil {
+		log.Printf("Error injecting server.go file: %v", err)
+		cobra.CheckErr(err)
+		return err
+	}
```
Ujstor, line 542: Create a directory under 'internal/logger' and place 'logger.go' there

cmd/template/framework/files/main/main.go.tmpl
```diff
@@ -2,14 +2,26 @@ package main
 
 import (
 	"fmt"
+	"log/slog"
```
Ujstor, line 5: There is also a fiber main.go, so the logger init needs to be included in this template as well

Ujstor (changes requested):
You are using 'log.info' only in 'main.go'. Could you include logging across all templates where it is appropriate? I thought that was the initial idea?

Documentation updates are also needed. You can add the new tree structure in 'index.md' and provide a brief explanation in 'creating-project.md'.

Please check why tests are not passing

mcheviron:
> You are using 'log.info' only in 'main.go'. Could you include logging across all templates where it is appropriate? I thought that was the initial idea?

OK. That's why I wanted to clarify what I was going to do in prev comment because I thought all of them use the same main.go template. I see now that fiber uses a different one, that's it or are there other templates to take into consideration?
> Documentation updates are also needed. You can add the new tree structure in 'index.md' and provide a brief explanation in 'creating-project.md'.

I'll add it. I just saw it

> Please check why tests are not passing

I didn't read the CI/CD pipeline so I'm still unaware of what it does. I'll take a look and if I'm stuck I'll ping you.

Cheers!

PS: Please ping me if i'm missing anything else. Thanks!

mcheviron:
@Ujstor Hi again. After an illegal amount of time I think I know why the CI/CD pipeline is angry at me:
```yml
jobs:
  install_dependencies:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Setup Go
        uses: actions/setup-go@v4
        with:
          go-version: "1.20"
      - name: Install Dependencies
        run: go mod download
  framework_matrix:
    needs: install_dependencies
    strategy:
      matrix:
        framework:
          [chi, gin, fiber, gorilla/mux, httprouter, standard-library, echo]
        driver:
          [mysql, postgres, sqlite, mongo, redis, none]
        goVersion: ["1.20"]
```

slog was experimental in 1.20 and the actual release is in 1.21

That's why I get this fuckery:
```
 Running [/home/runner/golangci-lint-1.55.2-linux-amd64/golangci-lint run --out-format=github-actions --path-prefix=chi --timeout=5m] in [/home/runner/work/go-blueprint/go-blueprint/chi] ...
  level=warning msg="[runner] Can't run linter goanalysis_metalinter: buildir: failed to load package : could not load export data: no export data for \"log/slog\""
  level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: buildir: failed to load package : could not load export data: no export data for \"log/slog\"\n\n"
  
  Error: golangci-lint exit with code 3
  Ran golangci-lint in 412ms
```

I think. I tried `act` on my local machine but it kind of worked so I'm not sure if I'm missing something. Anyways, what do you believe we should do?

Also  added the template to fiber as well. I just need to create a subdirectory in the internal directory and that's all, I believe. Cheers

Ujstor:
@mcheviron

I believe I may have misunderstood some things around implementation. I envision an implementation where, after a user creates a 'go-blueprint' project, a custom logger is automatically integrated into all generated template files (such as 'server.go,' 'database.go,' 'routes.go,' etc., depending on the frameworks, databases, or advanced options selected).

This approach would provide users with comprehensive examples of how the logger is utilized throughout the entire project. Consequently, users would gain a better understanding of its usage and be able to implement it more effectively in their own projects

Check cons in program.go and logic how new dir/subdir is created.

In generate-linter.yml, we can change the Go version, and the linter would not encounter any problems.

Sorry for delay,  I am little busy lately.

mcheviron:
> Sorry for delay, I am little busy lately.

I totally understand. I am as well. I think I have a better understanding of what you want and I think it's a good idea. I'll try to get into it as soon as possible and if I feel like I'm being too opinionated, i'll comment out the code but leave it there in case the user wishes to uncomment it.

Do you want me to update the linter myself? I believe updating the framework_matrix task would solve the issue but since it's the CI pipeline, I didn't want to touch it without bringing it up. Cheers

Ujstor:
Yeah, you can update the Go version in the linter, and that should do the trick.

Your logger is fine, use it in other packages and replace old logs with the new implementation. The current one is not ideal:

```go
```

Don't rush with code, there is another big PR that introduces websockets, and there are a lot of changes in templates. We can find the best way to implement it after  ws PR is merged.

Hit me up on Discord in DM; my username is the same, and we can discuss and find the best solution.

mcheviron:
> Hit me up on Discord in DM; my username is the same, and we can discuss and find the best solution.

Hi sorry for the delay. I tried pinging you but you can't be added via strangers.

> Yeah, you can update the Go version in the linter, and that should do the trick.
> 
> Your logger is fine, use it in other packages and replace old logs with the new implementation. The current one is not ideal:

I will generate some projects this weekend and reread our convo to make sure I put the logger in the right subdir and then use it throughout the projects. If you add me, i'll ping you there before I commit.

P.S: I'll update the linter version as well

mcheviron:
ci/cd is updated to use 1.22 while linting. it's better to sync the versions, 1.21.1 is used somewhere (that's just a suggestion, do what's best for the proj)

logger is put in its own pkg under internal

i added a Fatal method to the logger, it called log.Fatal under the hood. slog offers no API for fataling

if you want it to be used in the db initialisation for all the templates, we'd either do a dependancy injection on the New() for each db and use the logger therein. but since log.Fatal is used inside, I don't know if it's worth it. but it's doable

cmd/template/dbdriver/files/service/postgres.tmpl
```diff
@@ -28,26 +28,26 @@ var (
...
+func New() (Service, error) {
 	connStr := fmt.Sprintf("postgres://%s:%s@%s:%s/%s?sslmode=disable", username, password, host, port, database)
 	db, err := sql.Open("pgx", connStr)
 	if err != nil {
-		log.Fatal(err)
+		return nil, fmt.Errorf("error connecting to db: %w", err)
 	}
 	s := &service{db: db}
-	return s
+	return s, nil
```
H0llyW00dzZ, line 38: You don't have to set `nil` for this, because `nil` is already handled by the caller when you use it somewhere.

cmd/template/dbdriver/files/service/postgres.tmpl
```diff
@@ -28,26 +28,26 @@ var (
...
 
-func (s *service) Health() map[string]string {
+func (s *service) Health() (map[string]string, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
 	defer cancel()
 
 	err := s.db.PingContext(ctx)
 	if err != nil {
-		log.Fatalf(fmt.Sprintf("db down: %v", err))
+		return nil, fmt.Errorf("error pinging db: %w", err)
```
H0llyW00dzZ, line 47: For this one as well, it's important to use fatal.

cmd/template/dbdriver/files/service/postgres.tmpl
```diff
@@ -28,26 +28,26 @@ var (
...
 	}
 
 	return map[string]string{
 		"message": "It's healthy",
-	}
-}
+	}, nil
```
H0llyW00dzZ, line 52: For this one as well, you don't have to set `nil`.

H0llyW00dzZ (changes requested):
Request changed for this all database template it going wrong by putting nil and error

cmd/template/dbdriver/files/service/postgres.tmpl
mcheviron: > You don't have to set `nil` for this, because `nil` is already handled by the caller when you use it somewhere.

check with uj and his discussion with me. we don't want fatal's. this is Go not Java. The idea is to return errors and handle the errors via a tracing mechanism on the top level then terminating on certain errors and ultimately let the user choose what is fatal and what's not instead of greping for all fatals in the project like i did and changing them because that's lackluster in any production environment.

H0llyW00dzZ: > > You don't have to set `nil` for this, because `nil` is already handled by the caller when you use it somewhere.
> 
> check with uj and his discussion with me. we don't want fatal's. this is Go not Java. The idea is to return errors and handle the errors via a tracing mechanism on the top level then terminating on certain errors and ultimately let the user choose what is fatal and what's not instead of greping for all fatals in the project like i did and changing them because that's lackluster in any production environment.

It's not about Go or Java. The concern is about handling errors properly. Just imagine if you don't use fatal, then trying to connect to a database that is down could lead to unexpected behavior or crashes.

H0llyW00dzZ: the current implementation is already corrected using fatal and without `nil`

example:

https://github.com/Melkeydev/go-blueprint/blob/main/cmd/template/dbdriver/files/service/mysql.tmpl

H0llyW00dzZ: this how it work with fatal

```sh
$ go run cmd/api/main.go
2024/04/07 16:28:30 [H0llyW00dzZ Firewall] [INFO] Connecting to database: *********
2024/04/07 16:28:31 [H0llyW00dzZ Firewall] [FATAL] Failed to create firewall table: Error 1045 (28000): Access denied for user 'root'@'127.0.0.1' (using password: YES)
exit status 1
```

mcheviron: You don't fatal from inside a initialization process. You return an error and then handle it in the main function however you see fit. I see the merit of fataling whenever you see an initialization error. That was the discussion uj and I had and decided that the idiomatic way is to return an error (I don't know what you're on about nil honestly) and in the main function, we log the error and then terminate. This is more extensible because the user may want to retry etc. Also I don't understand what you're on about connecting to a database that's down. That's what errors are for mate. You connect to a database that's down, you get an error, you return it and handle it accordingly. We remained unopinionated by simply terminating like you showed in your example but via the error method the user may simply change this behavior to whatever she sees fitting. While fataling is a one way ticket. Thanks for your recommendations though for further info, again I suggest hitting up uj. Cheers mate


P.S: If your function can error, return the error. Don't expose a function signature that shows no errors only for the user realizing that it can blow up. Don't expect that each user will read your internal source code to know where panics will occur. That's literally why Go is an error based instead of exception based language.

H0llyW00dzZ: Also when it correct it show like this

```sh
$ go run cmd/api/main.go
2024/04/07 20:09:06 [H0llyW00dzZ Firewall] [INFO] Connecting to database: *********
2024/04/07 20:09:08 [H0llyW00dzZ Firewall] [INFO] Successfully initialized the firewall table.
2024/04/07 20:09:09 [H0llyW00dzZ Project] [INFO] Starting server on :8080

 ┌───────────────────────────────────────────────────┐ 
 │                H0llyW00dzZ Project                │ 
 │                   Fiber v2.52.4                   │ 
 │               http://127.0.0.1:8080               │ 
 │       (bound on host 0.0.0.0 and port 8080)       │ 
 │                                                   │ 
 │ Handlers ............ 539 Processes ........... 1 │ 
 │ Prefork ....... Disabled  PID ............. 28260 │ 
 └───────────────────────────────────────────────────┘ 
```

H0llyW00dzZ: And another this one is correct as well and follow idiomatic go practices

```go
	err := s.db.PingContext(ctx)
	if err != nil {
	}
```

mcheviron: Thanks for your input. I will wait for uj's review. Cheers

H0llyW00dzZ: > Thanks for your input. I will wait for uj's review. Cheers

Alright, good luck. Because The code is already correct, but the only thing you need to do is add a logger and structure it properly. I already know that the only thing missing in this repository is a logger.

H0llyW00dzZ: also another about `nil` you don't have to explicitly return `nil` because `nil` is already handled by the caller. For example, it's like this:

```go
package database

import (
	"context"
	"database/sql"

	"golang.org/x/crypto/bcrypt"
)

// User represents a user in the system with their basic credentials.
type User struct {
	ID       int64  // ID is the unique identifier for the user.
	Username string // Username is the user's username.
	Email    string // Email is the user's email address.
	// TODO: Implement this.
}

// ServiceAuthUser is an interface that defines methods for user authentication and management.
// It provides a contract for implementing user-related database operations.
type ServiceAuthUser interface {

	// ListUsers retrieves all users from the database.
	ListUsers(ctx context.Context) ([]User, error)

	// AuthenticateUser verifies user credentials against the stored credentials.
	AuthenticateUser(ctx context.Context, username, password string) (bool, error)
}

// serviceAuthUser is a concrete implementation of the ServiceAuthUser interface.
// It encapsulates the database connection and provides methods to interact with user data.
type serviceAuthUser struct {
	db *sql.DB
}

// NewServiceAuthUser creates a new instance of the ServiceAuthUser interface.
// It takes a database connection as a parameter and returns a new serviceAuthUser instance.
func NewServiceAuthUser(db *sql.DB) ServiceAuthUser {
	return &serviceAuthUser{db: db}
}

// ListUsers retrieves all users from the database.
// It executes a SELECT query to fetch user data and returns a slice of User structs.
func (s *serviceAuthUser) ListUsers(ctx context.Context) ([]User, error) {
	var users []User
	query := "SELECT id, username, email FROM users"
	rows, err := s.db.QueryContext(ctx, query)
	if err != nil {
		return nil, err
	}
	defer rows.Close()

	for rows.Next() {
		var user User
		if err := rows.Scan(&user.ID, &user.Username, &user.Email); err != nil {
			return nil, err
		}
		users = append(users, user)
	}

	if err := rows.Err(); err != nil {
		return nil, err
	}

	return users, nil
}

// AuthenticateUser checks if the provided username and password are correct.
// It queries the database for the user with the given username and compares the provided password with the stored hashed password.
// It returns true if authentication is successful, or false if not.
func (s *serviceAuthUser) AuthenticateUser(ctx context.Context, username, password string) (bool, error) {
	// Query the database for the user with the given username
	var passwordHash string
	err := s.db.QueryRowContext(ctx, "SELECT password_hash FROM users WHERE username = ?", username).Scan(&passwordHash)
	if err != nil {
		if err == sql.ErrNoRows {
			// User not found
			return false, nil
		}
		// An actual error occurred
		return false, err
	}

	// Compare the provided password with the hashed password from the database
	err = bcrypt.CompareHashAndPassword([]byte(passwordHash), []byte(password))
	if err != nil {
		// Password does not match
		return false, nil
	}

	// Password matches
	return true, nil
}
```

In combination with this:

- https://github.com/Melkeydev/go-blueprint/pull/209

H0llyW00dzZ: that's how idiomatic go best practices works

Melkeydev: > Also when it correct it show like this
> 
> ```shell
> $ go run cmd/api/main.go
> 2024/04/07 20:09:06 [H0llyW00dzZ Firewall] [INFO] Connecting to database: *********
> 2024/04/07 20:09:08 [H0llyW00dzZ Firewall] [INFO] Successfully initialized the firewall table.
> 2024/04/07 20:09:09 [H0llyW00dzZ Project] [INFO] Starting server on :8080
> 
>  ┌───────────────────────────────────────────────────┐ 
>  │                H0llyW00dzZ Project                │ 
>  │                   Fiber v2.52.4                   │ 
>  │               http://127.0.0.1:8080               │ 
>  │       (bound on host 0.0.0.0 and port 8080)       │ 
>  │                                                   │ 
>  │ Handlers ............ 539 Processes ........... 1 │ 
>  │ Prefork ....... Disabled  PID ............. 28260 │ 
>  └───────────────────────────────────────────────────┘ 
> ```

Hey - I thought I should chime in here because this great PR has definitely been out for too long without resolution. I discussed with the internal team, and we believe for go-blueprint that thing that makes sense is to throw Fatal if there is a DB connectivity error.

I am thinking about this from the perspective of a user, and if there is a an error connecting to the DB, but the app still spins up, it might be less than perfect user experience?

My intention is to not come in here and throw down a hammer, but again, asking for input on this. From my personal thinking, fatal on db error makes sense

H0llyW00dzZ: > > Also when it correct it show like this
> > ```shell
> > $ go run cmd/api/main.go
> > 2024/04/07 20:09:06 [H0llyW00dzZ Firewall] [INFO] Connecting to database: *********
> > 2024/04/07 20:09:08 [H0llyW00dzZ Firewall] [INFO] Successfully initialized the firewall table.
> > 2024/04/07 20:09:09 [H0llyW00dzZ Project] [INFO] Starting server on :8080
> > 
> >  ┌───────────────────────────────────────────────────┐ 
> >  │                H0llyW00dzZ Project                │ 
> >  │                   Fiber v2.52.4                   │ 
> >  │               http://127.0.0.1:8080               │ 
> >  │       (bound on host 0.0.0.0 and port 8080)       │ 
> >  │                                                   │ 
> >  │ Handlers ............ 539 Processes ........... 1 │ 
> >  │ Prefork ....... Disabled  PID ............. 28260 │ 
> >  └───────────────────────────────────────────────────┘ 
> > ```
> 
> Hey - I thought I should chime in here because this great PR has definitely been out for too long without resolution. I discussed with the internal team, and we believe for go-blueprint that thing that makes sense is to throw Fatal if there is a DB connectivity error.
> 
> I am thinking about this from the perspective of a user, and if there is a an error connecting to the DB, but the app still spins up, it might be less than perfect user experience?
> 
> My intention is to not come in here and throw down a hammer, but again, asking for input on this. From my personal thinking, fatal on db error makes sense

`I am thinking about this from the perspective of a user, and if there is a an error connecting to the DB, but the app still spins up, it might be less than perfect user experience?` For that, it's better to use `Fatal` for the database because, for the database, it's recommended to set it in `init` to initialize first before the app runs. So, when the database fails, the app won't run.

I have already implemented this, as shown in the example below:

```go
func init() {
	// Initialize the logger here with the necessary parameters.
	Logger.InitializeLogger("H0llyW00dzZ Firewall")
	// Initialize the context and cancel function
	schedulerCtx, cancelScheduler = context.WithCancel(context.Background())

	// Check if the necessary environment variables for the database connection are set.
	if os.Getenv(database.EnvMYSQLDBName) != "" && os.Getenv(database.EnvMYSQLDBPassword) != "" &&
		os.Getenv(database.EnvMYSQLDBUsername) != "" && os.Getenv(database.EnvMYSQLDBPort) != "" &&
		os.Getenv(database.EnvMYSQLDBHost) != "" {
		// Initialize the database service.
		db = database.New()

		// Initialize the firewall table if it doesn't exist.
		if err := createFirewallTable(); err != nil {
			Logger.LogFatalf("Failed to initialize firewall table: %v", err)
		} else {
			Logger.LogInfo("Successfully initialized the firewall table.")
		}
	}
}
```

also note that keeping fatal in `New` in this

```go
func New() Service {
	// Opening a driver typically will not attempt to connect to the database.
	db, err := sql.Open("mysql", fmt.Sprintf("%s:%s@tcp(%s:%s)/%s", username, password, host, port, dbname))
	if err != nil {
		// This will not be a connection error, but a DSN parse error or
		// another initialization error.
		log.Fatal(err)
	}
```

it will show the failure in `// Initialize the firewall table if it doesn't exist.` plus it won't making new connection because of this singleton pattern:

- #209

H0llyW00dzZ: Also, ignore the line `schedulerCtx, cancelScheduler = context.WithCancel(context.Background())` in example. That is for my worker goroutines.

H0llyW00dzZ: #### Another Example Method Using Recovery for Panic

```go
func init() {
    // Initialize the logger here with the necessary parameters.
    Logger.InitializeLogger("H0llyW00dzZ Firewall")
    // Initialize the context and cancel function
    schedulerCtx, cancelScheduler = context.WithCancel(context.Background())

    // Check if the necessary environment variables for the database connection are set.
    if os.Getenv(database.EnvMYSQLDBName) != "" && os.Getenv(database.EnvMYSQLDBPassword) != "" &&
        os.Getenv(database.EnvMYSQLDBUsername) != "" && os.Getenv(database.EnvMYSQLDBPort) != "" &&
        os.Getenv(database.EnvMYSQLDBHost) != "" {
        // Set up a defer to recover from any panic that might occur in database initialization.
        defer func() {
            if r := recover(); r != nil {
                Logger.LogFatalf("Failed to initialize the database: %v", r)
            }
        }()

        // Initialize the database service.
        db = database.New()

        // Initialize the firewall table if it doesn't exist.
        if err := createFirewallTable(); err != nil {
            Logger.LogFatalf("Failed to initialize firewall table: %v", err)
        } else {
            Logger.LogInfo("Successfully initialized the firewall table.")
        }
    }
}
```

It will show like this:

```sh
$ go run cmd/api/main.go
2024/04/10 08:36:54 [H0llyW00dzZ Firewall] [INFO] Initialization to database: *********
2024/04/10 08:36:55 [H0llyW00dzZ Firewall] [FATAL] Failed to initialize the database: Error 1045 (28000): Access denied for user 'root'@'127.0.0.1' (using password: YES)
exit status 1
```
```

Comment (H0llyW00dzZ) [reply cmd/template/dbdriver/files/service/postgres.tmpl:38]: And another issue that can occur during runtime (e.g., when a user is interacting with the database but the database goes down) can be covered by panic recovery in the middleware (e.g., if the error is not handled correctly).

That's it

**Review by H0llyW00dzZ — COMMENTED** [review]
This approach is better, as it avoids explicit `nil` and uses `panic` when initialization fails.

File: cmd/template/dbdriver/files/service/redis.tmpl [hunk 7]
```diff
@@ -46,21 +45,19 @@ func New() Service {
...
-
-  if result != "PONG" {
-    log.Fatalf(fmt.Sprintf("Unexpected ping response: %s", result))
-  }
 
 	return map[string]string{
 		"message": "It's healthy",
-	}
-}
+	}, nil
```
H0llyW00dzZ, line 62: This approach avoids explicit nil values for Redis and MySQL connections, which is beneficial for large codebases and scalability. It's better to remove explicit nil handling, as it will be handled by the caller elsewhere.

H0llyW00dzZ (changes requested):
by the way this still bad using `nil`

cmd/template/framework/files/logger.go.tmpl
```diff
@@ -0,0 +1,94 @@
...
+		if a.Key == "function" {
+			return slog.Attr{}
+		}
+
+		if config.AddSource && a.Key == slog.SourceKey {
+			source := a.Value.Any().(*slog.Source)
+			source.File = filepath.Base(source.File)
+		}
+
+		// if used in AWS or GCP (and Azure?), they'll add the timestamp themselves. Else uncomment to get timestamps
```
H0llyW00dzZ, line 44: It's better to enable this independently, even when used in AWS, GCP, or Azure, because the logger should be configured independently instead of depending on the `cloud provider` or whatever it is for timestamp handling.

H0llyW00dzZ (changes requested):
just letting you know again

Ujstor:
I am closing this repo, assuming there is no interest to continue implementation.

PR #192, opened 2024-02-25
Add websocket endpoint to advanced options

### LINKED ISSUE #174: Adding WebSocket

I think this blueprint lacks WebSocket options, which many people use. I believe at least the initialization of WebSockets or, if possible, simple functionality for creating rooms, connecting, disconnecting, broadcasting messages, and adding someone to a room would be helpful. Maybe something simpler that makes the code structure clear about what you're trying to achieve. I am new to it, and Melkey your blueprint pkg really improved my understanding of Golang very fast. Thanks!

### Disclaimer

- [X] I agree

MitchellBerend:

## Problem/Feature

Resolves #174 

## Description of Changes: 

This PR adds an option to the `--advanced` flag that let's users add a websocket endpoint to their `routes.go` file automatically.

#184 is still open so if that get's merged before this one, this pr will also include the flag option like that pr describes.

## Checklist

- [ ] I have self-reviewed the changes being requested

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -69,6 +69,7 @@ type Templater interface {
 	TestHandler() []byte
 	HtmxTemplRoutes() []byte
 	HtmxTemplImports() []byte
+	WebsocketImports() []byte
 }
 
 type DBDriverTemplater interface {
@@ -522,6 +523,14 @@ func (p *Project) CreateMainFile() error {
 		}
 	}
 
+	// if the websocket option is checked, a websocket dependency needs to
+	// be added to the routes depending on the framework choosen.
+	// Only fiber uses a different websocket library, the other frameworks
+	// all work with the same one
+	if p.AdvancedOptions["Websocket"] {
+		p.CreateWebsocketImports()
+	}
+
 	err = p.CreateFileWithInjection(internalServerPath, projectPath, "routes.go", "routes")
 	if err != nil {
 		log.Printf("Error injecting routes.go file: %v", err)
@@ -713,4 +722,23 @@ func (p *Project) CreateHtmxTemplates() {
 	}
 	p.AdvancedTemplates.TemplateRoutes = template.HTML(routeBuffer.String())
 	p.AdvancedTemplates.TemplateImports = template.HTML(importBuffer.String())
+}
+
+func (p *Project) CreateWebsocketImports() {
+	importsPlaceHolder := ""
+	if p.AdvancedOptions["Websocket"] {
+		importsPlaceHolder += string(p.FrameworkMap[p.ProjectType].templater.WebsocketImports())
+	}
+
+	importTmpl, err := template.New("imports").Parse(importsPlaceHolder)
+	if err != nil {
+		log.Fatal(err)
+	}
+	var importBuffer bytes.Buffer
+	err = importTmpl.Execute(&importBuffer, p)
+	if err != nil {
+		log.Fatal(err)
+	}
+	newImports := strings.Join([]string{string(p.AdvancedTemplates.TemplateImports), importBuffer.String()}, "\n")
+	p.AdvancedTemplates.TemplateImports = template.HTML(newImports)
 }
--- a/cmd/template/framework/files/routes/chi.go.tmpl
+++ b/cmd/template/framework/files/routes/chi.go.tmpl
@@ -19,6 +19,9 @@ func (s *Server) RegisterRoutes() http.Handler {
   {{if ne .DBDriver "none"}}
 	r.Get("/health", s.healthHandler)
   {{end}}
+  {{if .AdvancedOptions.Websocket}}
+	r.Get("/websocket", s.websocketHandler)
+  {{end}}
   {{.AdvancedTemplates.TemplateRoutes}}
 
 	return r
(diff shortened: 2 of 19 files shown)


MitchellBerend:
Talked to @Ujstor about the generated websocket handler and they made the fair suggestion of changing the generated handler to be more in line with what websockets are typically used for; That is that they should be sending data without interaction of the client beyond the initial connection request.

The initial idea is for the server to send a timestamp to the client every 0.5 seconds.

MitchellBerend:
During hand testing this feature the `fiber` option did not generate a full project [^1]. The linting jobs did pass but I think this qualifies as a 'flaky' test to me. Maybe these tests need to also include a build test of the generated project to make sure the application stays functional.

[^1]: the `<-` got sanitized into

README.md
```diff
@@ -94,22 +96,24 @@ Choose from a variety of supported database drivers:
...
 
-Blueprint is focused on being as minimalistic as possible. That being said, we wanted to offer the ability to add other features people may want without bloating the overall experience. 
+Blueprint is focused on being as minimalistic as possible. That being said, we wanted to offer the ability to add other features people may want without bloating the overall experience.
 
 You can now use the `--advanced` flag when running the `create` command to get access to the following features. This is a multi-option prompt; one or more features can be used at the same time:
 
 - [HTMX](https://htmx.org/) support using [Templ](https://templ.guide/)
 - CI/CD workflow setup using [Github Actions](https://docs.github.com/en/actions)
-
+- [Websocket](https://en.wikipedia.org/wiki/WebSocket) sets up a websocket endpoint
```
Melkeydev, line 113: I think this needs to either link to some Go related source material and not just a wikipedia article on web sockets

cmd/program/program.go
```diff
@@ -711,6 +720,38 @@ func (p *Project) CreateHtmxTemplates() {
...
+	// have to here
+	err := utils.GoGetPackage(appDir, websocketDependency)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	importsPlaceHolder := ""
+	if p.AdvancedOptions["Websocket"] {
+		importsPlaceHolder += string(p.FrameworkMap[p.ProjectType].templater.WebsocketImports())
+	}
```
Melkeydev, line 744: Why would this be necessary if this function is only called when the use selects web socket as an option?
Can we omit the if?

cmd/program/program.go
```diff
@@ -711,6 +720,38 @@ func (p *Project) CreateHtmxTemplates() {
...
+
+	importTmpl, err := template.New("imports").Parse(importsPlaceHolder)
+	if err != nil {
+		log.Fatal(err)
```
Melkeydev, line 748: Nit: I want to fix the error handling in go-blueprint.
Can we do something like
```
if err != nil {
    log.Fatalf("function failed: %v", err)
}
```

cmd/program/program.go
```diff
@@ -711,6 +720,38 @@ func (p *Project) CreateHtmxTemplates() {
...
+	}
+	var importBuffer bytes.Buffer
+	err = importTmpl.Execute(&importBuffer, p)
+	if err != nil {
+		log.Fatal(err)
+	}
```
Melkeydev, line 754: Same here

cmd/template/framework/files/routes/chi.go.tmpl
```diff
@@ -42,3 +49,30 @@ func (s *Server) healthHandler(w http.ResponseWriter, r *http.Request) {
 	_, _ = w.Write(jsonResp)
 }
 {{end}}
+
+{{if .AdvancedOptions.Websocket}}
+func (s *Server) websocketHandler(w http.ResponseWriter, r *http.Request) {
+	socket, err := websocket.Accept(w, r, nil)
+
+	if err != nil {
+		log.Print("could not open websocket")
```
Melkeydev, line 58: We need to surface the error to understand what the error is

cmd/template/framework/files/routes/echo.go.tmpl
```diff
@@ -34,3 +42,33 @@ func (s *Server) healthHandler(c echo.Context) error {
...
+func (s *Server) websocketHandler(c echo.Context) error {
+	w := c.Response().Writer
+	r := c.Request()
+	socket, err := websocket.Accept(w, r, nil)
+
+	if err != nil {
+		log.Print("could not open websocket")
+		_, _ = w.Write([]byte("could not open websocket"))
+		w.WriteHeader(http.StatusInternalServerError)
+		return nil
```
Melkeydev, line 56: Same here - we need to surface the error

Melkeydev (commented):
Great PR - excited for this.
A few comments and i think better error handling is needed

cmd/program/program.go
MitchellBerend: Ye good point, I think this is a left over if statement from the refactor I did before sending this up.

README.md
MitchellBerend: I could link the most used websocket library [^1] in the generated code but that would not encompass fiber as an option as that is using it's own websocket implementation.

Is that okay with you?

[^1]: https://pkg.go.dev/nhooyr.io/websocket

Ujstor (changes requested):
@MitchellBerend 

PR #184  is merged and introduced an inside flag feature. Some minor changes are needed so that the WS flag can be used.

cmd/template/framework/files/routes/fiber.go.tmpl
```diff
@@ -27,3 +36,34 @@ func (s *FiberServer) healthHandler(c *fiber.Ctx) error {
...
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			payload := fmt.Sprintf("server timestamp: %d", time.Now().UnixNano())
+			if err := con.WriteMessage(websocket.TextMessage, []byte(payload)); err != nil {
+				log.Printf("could not write to socket: %v", e)
```
briancbarrow, line 62:
```suggestion
				log.Printf("could not write to socket: %v", err)
```

briancbarrow (changes requested):
Just that one change with the err and then it should be good to go.

cmd/template/framework/files/routes/fiber.go.tmpl
MitchellBerend: I don't know how I didn't catch this but I also think ci should have caught a non building option.
I created [an issue](https://github.com/Melkeydev/go-blueprint/issues/195) about that which I plan to also fix.

Ujstor (approved):
Nicely done! GG

PR #206, opened 2024-03-30
Implement [Fiber Framework] Middleware Rate Limiter

H0llyW00dzZ:
By submitting this pull request, I confirm that my contribution is made under the terms of the MIT license.

## Problem/Feature

Implementing this rate limiter is pretty useful. I've been using it on my large RESTful APIs (500+ Handlers) with HTMX (for front end).

## Description of Changes: 

- [+] feat(program.go): add middleware limiter to fiber package
- [+] feat(fiber.go.tmpl): add setRouteRateLimiter function to set rate limit configuration

## Checklist

- [X] I have self-reviewed the changes being requested
- [ ] I have updated the documentation (if applicable)

PR #209, opened 2024-04-01
[Database Template] Implement Singleton Pattern

H0llyW00dzZ:

## Problem/Feature

This pull request implements the Singleton pattern for database connections, which is useful for reusing connections efficiently.

## Description of Changes: 

- [+] feat(mysql.tmpl): implement singleton pattern for dbInstance in New function
- [+] fix(mysql.tmpl): correct dbInstance assignment syntax

Note: This a fix syntax for example it can be used multipe services such as
```go
dbInstance = &service{
		db:       db,
		authuser: NewServiceAuthUser(db), // Initialize the auth user service
	}
```

## Checklist

### DIFF
--- a/cmd/template/dbdriver/files/service/mongo.tmpl
+++ b/cmd/template/dbdriver/files/service/mongo.tmpl
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"sync"
 	"time"
 
 	"go.mongodb.org/mongo-driver/mongo"
@@ -25,21 +24,18 @@ var (
 	host     = os.Getenv("DB_HOST")
 	port     = os.Getenv("DB_PORT")
 	//database = os.Getenv("DB_DATABASE")
-	dbInstance *service
-	once       sync.Once
 )
 
 func New() Service {
-	// Note: This is a different method for MongoDB compared to MySQL because, unlike MongoDB, you don't need to set `sync.Once` for MySQL.
-	once.Do(func() {
-		client, err := mongo.Connect(context.Background(), options.Client().ApplyURI(fmt.Sprintf("mongodb://%s:%s", host, port)))
+	client, err := mongo.Connect(context.Background(), options.Client().ApplyURI(fmt.Sprintf("mongodb://%s:%s", host, port)))
 
-		if err != nil {
-			log.Fatal(err)
-		}
-		dbInstance = &service{db: client}
-	})
-	return dbInstance
+	if err != nil {
+		log.Fatal(err)
+
+	}
+	return &service{
+		db: client,
+	}
 }
 
 func (s *service) Health() map[string]string {
--- a/cmd/template/dbdriver/files/service/mongo.tmpl
+++ b/cmd/template/dbdriver/files/service/mongo.tmpl
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"sync"
 	"time"
 
 	"go.mongodb.org/mongo-driver/mongo"
@@ -24,18 +25,21 @@ var (
 	host     = os.Getenv("DB_HOST")
 	port     = os.Getenv("DB_PORT")
 	//database = os.Getenv("DB_DATABASE")
+	dbInstance *service
+	once       sync.Once
 )
 
 func New() Service {
-	client, err := mongo.Connect(context.Background(), options.Client().ApplyURI(fmt.Sprintf("mongodb://%s:%s", host, port)))
+	// Note: This is a different method for MongoDB compared to MySQL because, unlike MongoDB, you don't need to set `sync.Once` for MySQL.
+	once.Do(func() {
+		client, err := mongo.Connect(context.Background(), options.Client().ApplyURI(fmt.Sprintf("mongodb://%s:%s", host, port)))
 
-	if err != nil {
-		log.Fatal(err)
-
-	}
-	return &service{
-		db: client,
-	}
+		if err != nil {
+			log.Fatal(err)
+		}
+		dbInstance = &service{db: client}
+	})
+	return dbInstance
 }
 
 func (s *service) Health() map[string]string {
--- a/cmd/template/dbdriver/files/service/mysql.tmpl
+++ b/cmd/template/dbdriver/files/service/mysql.tmpl
@@ -21,14 +21,20 @@ type service struct {
 }
 
 var (
-	dbname   = os.Getenv("DB_DATABASE")
-	password = os.Getenv("DB_PASSWORD")
-	username = os.Getenv("DB_USERNAME")
-	port     = os.Getenv("DB_PORT")
-	host     = os.Getenv("DB_HOST")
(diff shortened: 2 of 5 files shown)


H0llyW00dzZ:
tested live in production:

image

json:

```json
{"status":"up","open_connections":"2","idle":"1","max_lifetime_closed":"0","message":"It's healthy","in_use":"1","wait_count":"0","wait_duration":"0s","max_idle_closed":"0"}
```

briancbarrow (changes requested):
@H0llyW00dzZ this is a good idea, but again we'd need the change applied to all database options.

H0llyW00dzZ:
@briancbarrow how about that MongoDB I've been Implement it ?

H0llyW00dzZ:
btw the ci error lmao

image

MitchellBerend:
@H0llyW00dzZ I peaked at your action results real quick (for both #209 and #206) and github seems to be having issues right now.

https://www.githubstatus.com/history

H0llyW00dzZ:
> @H0llyW00dzZ I peaked at your action results real quick (for both #209 and #206) and github seems to be having issues right now.
> 
> https://www.githubstatus.com/history

@MitchellBerend it's unstable, look this a weird failure

image

cmd/template/dbdriver/files/service/mongo.tmpl
```diff
@@ -24,18 +25,21 @@ var (
...
-	client, err := mongo.Connect(context.Background(), options.Client().ApplyURI(fmt.Sprintf("mongodb://%s:%s", host, port)))
-
-	if err != nil {
-		log.Fatal(err)
-
-	}
-	return &service{
-		db: client,
-	}
+	// Note: This is a different method for MongoDB compared to MySQL because, unlike MongoDB, you don't need to set `sync.Once` for MySQL.
```
H0llyW00dzZ, line 33: > [!NOTE]
> Regarding MongoDB, I think it does not support connecting to multiple databases in a single connection like MySQL. In MySQL, you can connect to multiple databases using a single connection with a singleton pattern.

H0llyW00dzZ (commented):
# INFO

H0llyW00dzZ:
> It seems that GitHub is having issues with actions. I reran manually, and more checks are passing
> 
> https://www.githubstatus.com/history

this a weird

image

H0llyW00dzZ:
now this one error again lmao

image

MitchellBerend:
The cache posting issue is resolved in #213 since that removes that action from ci, not sure if this issue is related to yesterday's outage but maybe it's safer to wait for that pr to be merged.

H0llyW00dzZ:
> The cache posting issue is resolved in #213 since that removes that action from ci, not sure if this issue is related to yesterday's outage but maybe it's safer to wait for that pr to be merged.

@MitchellBerend The cache posting issue seems to be a race condition because Node.js cannot catch it. Also I don't recommend running CI/CD on Node.js, which is an interpreted language, especially for running lots of CI tasks like this. Unlike compiled languages that can run directly on the machine, such as Go, Node.js adds an extra layer of interpretation.

Ujstor:
By the end of the weekend, linter fix will be merged

Ujstor:
Updated workflow will be used on a new commit. I can't run it manually.

H0llyW00dzZ:
> Updated workflow will be used on a new commit. I can't run it manually.

@Ujstor this singleton pattern supports Fiber storage as well and provides perfect scalability for rate limiting.

- #206

For example:

```go
package database

import (
	"context"
	"database/sql"
	"fmt"
	"log"
	"os"
	"time"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/storage/mysql/v2"

	_ "github.com/go-sql-driver/mysql"
	_ "github.com/joho/godotenv/autoload"
)

type Service interface {
	Health() map[string]string
	// RateLimiter returns the [fiber.Storage] interface for rate limiting.
	RateLimiter() fiber.Storage
}

type service struct {
	db          *sql.DB
	ratelimiter fiber.Storage
}

var (
	dbname     = os.Getenv("DB_DATABASE")
	password   = os.Getenv("DB_PASSWORD")
	username   = os.Getenv("DB_USERNAME")
	port       = os.Getenv("DB_PORT")
	host       = os.Getenv("DB_HOST")
	dbInstance *service
)

func New() Service {
	// Reuse Connection
	if dbInstance != nil {
		return dbInstance
	}

	// Opening a driver typically will not attempt to connect to the database.
	db, err := sql.Open("mysql", fmt.Sprintf("%s:%s@tcp(%s:%s)/%s", username, password, host, port, dbname))
	if err != nil {
		// This will not be a connection error, but a DSN parse error or
		// another initialization error.
		log.Fatal(err)
	}
	db.SetConnMaxLifetime(0)
	db.SetMaxIdleConns(50)
	db.SetMaxOpenConns(50)

	dbInstance = &service{
		db: db,
		ratelimiter: mysql.New(mysql.Config{
			Db:         db,
			Reset:      false,
			GCInterval: 10 * time.Second,
			Table:      "rate_limiter",
		}), // Initialize the rate limiter storage
	}
	return dbInstance
}

func (s *service) Health() map[string]string {
	defer cancel()

	err := s.db.PingContext(ctx)
	if err != nil {
	}

	return map[string]string{
		"message": "It's healthy",
	}
}

// RateLimiter returns the [fiber.Storage] interface for rate limiting.
func (s *service) RateLimiter() fiber.Storage {
	return s.ratelimiter
}
```

In the database:

image

H0llyW00dzZ:
> @H0llyW00dzZ this is a good idea, but again we'd need the change applied to all database options.

@briancbarrow all done

briancbarrow (approved):

H0llyW00dzZ:
@briancbarrow by the way, for `Redis` and `MongoDB`, might not have to set a singleton pattern, I think, because they are both mostly used for caching. Using `sync.Once` might cause issues because, unlike other databases like `MySQL`, `PostgreSQL`, and `SQLite`, which are stable and can recover from connection issues (I've tested this by killing connections and database servers, and the Go routines can reconnect without using previous connection).

briancbarrow:
> @briancbarrow by the way, for `Redis` and `MongoDB`, might not have to set a singleton pattern, I think, because they are both mostly used for caching. Using `sync.Once` might cause issues because, unlike other databases like `MySQL`, `PostgreSQL`, and `SQLite`, which are stable and can recover from connection issues (I've tested this by killing connections and database servers, and the Go routines can reconnect without using previous connection).
> 

Okay. Will you remove that code from the Redis and Mongo files?

H0llyW00dzZ:
@briancbarrow done

H0llyW00dzZ:
Also, note that the singleton pattern is more stable and better suited for traditional SQL databases (e.g., `MySQL`, `PostgreSQL`, `SQLite`) compared to using a pooling pattern. On the other hand, the pooling pattern is more stable and better suited for NoSQL databases such as `Redis` and `MongoDB`, which are primarily used for caching.

Issue #186, opened 2024-02-17
[Feature Request] Adding Tailwind Configuration to the Initial Setup


Nevoral:
### Tell us about your feature request

Right now, we have everything ready for web development, but Tailwind is part of my stack, and I need to configure it manually. So, adding a Tailwind option for the initial setup should be fine.

### Disclaimer

- [X] I agree

briancbarrow:
@Nevoral I agree this would be a great addition. It has been on my radar as well. At some point I plan on adding it, but if you want to tackle adding it to the advanced flag process, let me know and I will assign the issue to you.

briancbarrow:
Awesome. I've assigned the issue to you @Nevoral.

PR #221, opened 2024-04-07
move js/ into assets/js/ and make assets/ as static file server

kobamkode:

## Problem/Feature

Move `js/` into `assets/js` and make `assets/` as a static file server, so user can easily add a custom folder like `cmd/web/assets/css/` and link it into `base.templ` if they want it. 

## Description of Changes: 

- modify `cmd/program/program.go` to create assets/ 
- modify htmx src in `cmd/template/advanced/files/htmx/base.templ.tmpl`
- modify `go:embed "js"`  to `go:embed "assets"`
- modify frameworks static file routes in `cmd/template/advanced/files/htmx/routes`
- modify some file tree in docs

## Checklist

### DIFF
--- a/cmd/template/advanced/files/htmx/efs.go.tmpl
+++ b/cmd/template/advanced/files/htmx/efs.go.tmpl
@@ -2,5 +2,5 @@ package web
 
 import "embed"
 
-//go:embed "js"
+//go:embed "assets"
 var Files embed.FS
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -426,19 +426,19 @@ func (p *Project) CreateMainFile() error {
 			return err
 		}
 
-		err = os.Mkdir(fmt.Sprintf("%s/%s/js", projectPath, cmdWebPath), 0755)
+		err = os.MkdirAll(fmt.Sprintf("%s/%s/assets/js", projectPath, cmdWebPath), 0755)
 		if err != nil {
 			cobra.CheckErr(err)
 		}
 
-		htmxMinJsFile, err := os.Create(fmt.Sprintf("%s/%s/js/htmx.min.js", projectPath, cmdWebPath))
+		htmxMinJsFile, err := os.Create(fmt.Sprintf("%s/%s/assets/js/htmx.min.js", projectPath, cmdWebPath))
 		if err != nil {
 			cobra.CheckErr(err)
 		}
 		defer htmxMinJsFile.Close()
 
 		htmxMinJsTemplate := advanced.HtmxJSTemplate()
-		err = os.WriteFile(fmt.Sprintf("%s/%s/js/htmx.min.js", projectPath, cmdWebPath), htmxMinJsTemplate, 0644)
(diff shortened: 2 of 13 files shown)


H0llyW00dzZ:
it's better to made it own framework I think hahahaha

cmd/template/advanced/files/htmx/efs.go.tmpl
```diff
@@ -2,5 +2,5 @@ package web
 
 import "embed"
 
-//go:embed "js"
+//go:embed "assets"
```
EssaAlshammri, line 5: doesn't this needs to be assets/*

kobamkode: @H0llyW00dzZ, it sounds great!, but for now the reason is to help user who wants to add a simple custom css or add some images with just putting their folder into `assets/` and load it in their templ. :heart: 
  
@EssaAlshammri, I've test it with generating all the go-blueprint supported frameworks + htmx and the htmx is still loads and works. For your suggestion, I think I need to test it more further. Thx for reviewing my code btw. :+1:

kobamkode: > doesn't this needs to be assets/*

@EssaAlshammri 

I've added more test stuff in file server like css file and img file with `go:embed "assets"` directive, it looks fine. Then I reread the embed pkg documentation, interestingly I found:

> If a pattern names a directory, all files in the subtree rooted at that directory are embedded (recursively), except that files with names beginning with ‘.’ or ‘_’ are excluded.

Should I update to `assets/*` ?

H0llyW00dzZ: > @H0llyW00dzZ, it sounds great!, but for now the reason is to help user who wants to add a simple custom css or add some images with just putting their folder into `assets/` and load it in their templ. ❤️
> 
> @EssaAlshammri, I've test it with generating all the go-blueprint supported frameworks + htmx and the htmx is still loads and works. For your suggestion, I think I need to test it more further. Thx for reviewing my code btw. 👍

@kobamkode
How can you design the CSS for htmx? I am so bad at designing websites, to be honest.

EssaAlshammri: yeah the way you did it was right

I'm sorry it didn't work for me at first and worked when added * 

but now everything is working as expected 

yeah I'm using it for tailwind styles

image

briancbarrow (approved):

EssaAlshammri:
There should have been a css folder with a default css file to be easier to add styles to the web app

Or better add tailwind with the advance flag

H0llyW00dzZ:
> There should have been a css folder with a default css file to be easier to add styles to the web app
> 
> Or better add tailwind with the advance flag

I agree with the suggestion for Tailwind, as mentioned in #186.

Because Go + HTMX is better, and the only missing component is CSS, unlike other languages, especially Java, which consumes a lot of memory.

Also I don't understand why most developers write Java when it's better to write Go, especially for microservices.

PR #233, opened 2024-05-15
Add Tailwind to advance options.

### LINKED ISSUE #186: [Feature Request] Adding Tailwind Configuration to the Initial Setup

PrajvalBadiger:

## Problem/Feature
Resolves #186 

## Description of Changes: 
This PR adds an option to the `--advanced` flag that let's users add Tailwind CSS framework to the project
- This change creates package.json, tailwind.config.js, input.css and output.css files
- pnpm is used to install tailwind via package.json as I couldn't test with npm
- updated Makefile to run tailwind for the build recipe. which is also used in .air.toml

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ You can now use the `--advanced` flag when running the `create` command to get a
 - [HTMX](https://htmx.org/) support using [Templ](https://templ.guide/)
 - CI/CD workflow setup using [Github Actions](https://docs.github.com/en/actions)
 - [Websocket](https://pkg.go.dev/nhooyr.io/websocket) sets up a websocket endpoint
+- [Tailwind](https://tailwindcss.com/) Css framework
 
 <a id="blueprint-ui"></a>
 
@@ -165,9 +166,14 @@ For the websocket:
 go-blueprint create --advanced --feature websocket
 ```
 
+For Tailwind:
+```bash
+go-blueprint create --advanced --feature tailwind
+```
+
 Or all features at once:
 ```bash
-go-blueprint create --name my-project --framework chi --driver mysql --advanced --feature htmx --feature githubaction --feature websocket
+go-blueprint create --name my-project --framework chi --driver mysql --advanced --feature htmx --feature githubaction --feature websocket --feature tailwind
 ```
 
 <p align="center">
--- a/README.md
+++ b/README.md
@@ -113,6 +113,8 @@ You can now use the `--advanced` flag when running the `create` command to get a
 - [Websocket](https://pkg.go.dev/nhooyr.io/websocket) sets up a websocket endpoint
 - [Tailwind](https://tailwindcss.com/) Css framework
 
+Note: selecting tailwind option automatically selects htmx.
+
 <a id="blueprint-ui"></a>
 
 <h2>
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -247,7 +247,7 @@ var createCmd = &cobra.Command{
 		if options.Advanced.Choices["Tailwind"] {
 			options.Advanced.Choices["Htmx"] = true
 			fmt.Println(endingMsgStyle.Render("• Install the tailwind standalone cli if you haven't already, grab the executable for your platform from the latest release on GitHub\n"))
-			fmt.Println(endingMsgStyle.Render("• Tailwind GitHub: https://github.com/tailwindlabs/tailwindcss/releases/\n"))
+			fmt.Println(endingMsgStyle.Render("• More info about the Tailwind CLI: https://tailwindcss.com/blog/standalone-cli\n"))
 		}
 
 		if options.Advanced.Choices["Htmx"] {
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -244,6 +244,12 @@ var createCmd = &cobra.Command{
 		fmt.Println(endingMsgStyle.Render("\nNext steps:"))
 		fmt.Println(endingMsgStyle.Render(fmt.Sprintf("• cd into the newly created project with: `cd %s`\n", project.ProjectName)))
 
+		if options.Advanced.Choices["Tailwind"] {
+			options.Advanced.Choices["Htmx"] = true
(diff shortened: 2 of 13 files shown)


briancbarrow:
@PrajvalBadiger Thanks for submitting this PR. I wanted to ask what you thought about changing the implementation to use the [Tailwind Standalone CLI](https://tailwindcss.com/blog/standalone-cli) instead of installing the package with Node.js? 

I've used the standalone CLI in my projects and it works well. We'd have to add some language to the output saying to make sure they have it installed, but we are already doing something similar with Templ so I don't think that will be an issue. 

The other thing we probably want to add is some checks to make sure that the tailwind option is not used without the htmx option.

PrajvalBadiger:
> @PrajvalBadiger Thanks for submitting this PR. I wanted to ask what you thought about changing the implementation to use the [Tailwind Standalone CLI](https://tailwindcss.com/blog/standalone-cli) instead of installing the package with Node.js?
> 
> I've used the standalone CLI in my projects and it works well. We'd have to add some language to the output saying to make sure they have it installed, but we are already doing something similar with Templ so I don't think that will be an issue.
> 

@briancbarrow, Thanks for the review. I was not aware of the Tailwind Standalone CLI and had used Node.js for Tailwind in my projects. A huge dependency on Node.js would be reduced by using the standalone CLI.

> The other thing we probably want to add is some checks to make sure that the tailwind option is not used without the htmx option.

There a few ways we can implement this:
1. By Default selecting the tailwind option is disabled and user can only select tailwind option if htmx option is selected
2. If tailwind option is selected, htmx is automatically selected
3. If only the tailwind option is used without htmx option, the program fails and print an error. If this is the case the user has to redo all the steps and select htmx option with tailwind option.

briancbarrow:
> There a few ways we can implement this:

> By Default selecting the tailwind option is disabled and user can only select tailwind option if htmx option is selected
> If tailwind option is selected, htmx is automatically selected
> If only the tailwind option is used without htmx option, the program fails and print an error. If this is the case the user has to redo all the steps and select htmx option with tailwind option.

I would be good with either the first or second option.

PrajvalBadiger:
1. I have implemented the second option - If tailwind option is selected, htmx is automatically selected.
2. Node.js dependency is droped and tailwind standalone cli can be used now.
3. Added note to remind users to install tailwind standalone cli.

cmd/create.go
```diff
@@ -244,6 +244,12 @@ var createCmd = &cobra.Command{
 		fmt.Println(endingMsgStyle.Render("\nNext steps:"))
 		fmt.Println(endingMsgStyle.Render(fmt.Sprintf("• cd into the newly created project with: `cd %s`\n", project.ProjectName)))
 
+		if options.Advanced.Choices["Tailwind"] {
+			options.Advanced.Choices["Htmx"] = true
+			fmt.Println(endingMsgStyle.Render("• Install the tailwind standalone cli if you haven't already, grab the executable for your platform from the latest release on GitHub\n"))
+			fmt.Println(endingMsgStyle.Render("• Tailwind GitHub: https://github.com/tailwindlabs/tailwindcss/releases/\n"))
```
briancbarrow, line 250: I would change this to point to the page explaining the Standalone CLI. That page will give more info about it and how to install it. It is still a relatively unknown feature so pointing people to an explainer page will be helpful. 
```suggestion
			fmt.Println(endingMsgStyle.Render("• More info about the Tailwind CLI: https://tailwindcss.com/blog/standalone-cli\n"))
```

cmd/steps/steps.go
```diff
@@ -108,6 +108,11 @@ func InitSteps(projectType flags.Framework, databaseType flags.Database) *Steps
 						Title: "Websocket endpoint",
 						Desc:  "Add a websocket endpoint",
 					},
+					{
+						Flag:  "Tailwind",
+						Title: "TailwindCSS",
+						Desc:  "A utility-first CSS framework",
```
briancbarrow, line 114:
```suggestion
						Desc:  "A utility-first CSS framework (selecting this will automatically add HTMX)",
```

briancbarrow (changes requested):
This is looking great. Some small changes to make things clearer in the output of the cli.

briancbarrow:
Thanks @PrajvalBadiger! The only thing left is it looks like there are commits on here that are from previously merged PRs. Can you do a rebase and only keep the commits that are relevant to this PR?

README.md
```diff
@@ -111,6 +111,9 @@ You can now use the `--advanced` flag when running the `create` command to get a
 - [HTMX](https://htmx.org/) support using [Templ](https://templ.guide/)
 - CI/CD workflow setup using [Github Actions](https://docs.github.com/en/actions)
 - [Websocket](https://pkg.go.dev/nhooyr.io/websocket) sets up a websocket endpoint
+- [Tailwind](https://tailwindcss.com/) Css framework
+
+Note: selecting tailwind option automatically selects htmx.
```
Melkeydev, line 116: nice

cmd/program/program.go
```diff
@@ -216,7 +216,7 @@ func (p *Project) CreateMainFile() error {
 	// check if AbsolutePath exists
 	if _, err := os.Stat(p.AbsolutePath); os.IsNotExist(err) {
 		// create directory
-		if err := os.Mkdir(p.AbsolutePath, 0754); err != nil {
+		if err := os.Mkdir(p.AbsolutePath, 0o754); err != nil {
```
Melkeydev, line 219: What is this doing?

briancbarrow (approved):
LGTM

cmd/program/program.go
PrajvalBadiger: This is nothing to do with the PR itself, go lsp in neovim auto formats octals 0o.

PR #251, opened 2024-06-07
Add flake for Nix users

Kodlak15:

## Problem/Feature

Flakes are a Nix solution that allow for reproducible package builds, development environments, and more. This flake would allow Nix users to use go-blueprint in a way that better interfaced with the rest of their system. 

## Description of Changes: 

- Create flake.nix file and include a package that can be used by Nix users to run go-blueprint or add it to their system

## Checklist

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,29 @@
+{
+  pkgs,
+  lib,
+}: let
+  go-blueprint = pkgs.buildGoModule {
+    pname = "go-blueprint";
+    version = "0.5.14";
+    src = ./.;
+
+    # Needs to be updated each time dependencies change
+    # Use lib.fakeHash and run 'nix build .#' to obtain the new hash
+    # vendorHash = lib.fakeHash; # uncomment this and comment the definition below out to obtain the new hash
+    vendorHash = "sha256-WBzToupC1/O70OYHbKk7S73OEe7XRLAAbY5NoLL7xvw=";
+
+    meta = with lib; {
+      description = "The ultimate golang blueprint library";
+      homepage = "https://github.com/Melkeydev/go-blueprint";
+      licence = licenses.mit;
+    };
+  };
+in
+  # Output a shell script that runs go-blueprint with go as a runtime dependency
+  pkgs.writeShellApplication {
+    name = "go-blueprint";
+    runtimeInputs = with pkgs; [go];
+    text = ''
+      "${go-blueprint}/bin/go-blueprint" "$@"
+    '';
+  }
--- a/default.nix
+++ b/default.nix
@@ -3,8 +3,7 @@
   lib,
 }: let
   go-blueprint = pkgs.buildGoModule {
-    pname = "go-blueprint";
-    version = "0.5.14";
+    name = "go-blueprint";
     src = ./.;
 
     # Needs to be updated each time dependencies change
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,317 @@
+{
+  "nodes": {
+    "devshell": {
+      "inputs": {
+        "flake-utils": "flake-utils_2",
+        "nixpkgs": [
+          "nixvim",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1717408969,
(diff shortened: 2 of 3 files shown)


Ujstor:
Is there a way to update flake to the new release? Currently, the version is hardcoded. I am thinking of an action that is triggered by a Go release. When it finishes, it can create/update the flake?

```yml
name: flake.nix

on:
  workflow_run:
    workflows: ["goreleaser"]
    types:
      - completed
jobs:
...
```

Kodlak15:
So, I initially created this package so that I could just use it on my own system, thus the use of fetchFromGitHub. Since the flake resides within the repository itself in this context, my thought was I could simply remove fetchFromGitHub as well as the version specification and use `src = ./.` instead. If my understanding is correct, this should just run/build the most recently published version of go-blueprint to the main branch.

Kodlak15:
Also worth noting, a minor gripe I have with this setup is the fact that adding the package to your NixOS or home-manager configurations makes the Go binary from Nixpkgs available globally. It is typically best practice not to have global installations of programming languages, rather one should create local shells that contain development/build dependencies. In this program, Go is needed in the PATH when the program is run to generate the go.mod file. The only solution I have been able to conjure up for this issue is to link the Go binary to the packages output directory like so:

```
...
          pkgs.symlinkJoin rec {
            name = "go-blueprint";
            # Go needed in PATH to generate go.mod
            paths = [module] ++ [pkgs.go];
            buildInputs = [pkgs.makeWrapper];
            postBuild = "wrapProgram $out/bin/${name} --prefix PATH : $out/bin";
          };
...
```

This is a non-issue if one simply wants to run the package directly via the flake with `nix run github:Kodlak15/go-blueprint` or `nix run github:Melkeydev/go-blueprint` (assuming this were to be merged). I am open to suggestions if anyone knows of a better way to handle this, but for the time being this is what I have had to do to make things work.

Ujstor:
That a thing, I don't know what the best practices are. I was able to build it locally. I am familiar with the concept, but Nix isn't my primary daily driver. We can integrate it, and I can see the benefits. If people will use Blueprint with the Nix package manager, why not

We need Nix experts to guide us in the right direction

Kodlak15:
Yeah I agree regarding that last point. I am by no means a "Nix expert". I have been daily driving NixOS for the past 6 months or so and have learned a thing or two, but it is a rather deep rabbit hole to dive into. I also recognize that this solves a problem for a rather small subset of users, but I figured it could be helpful for whatever small proportion of this projects users that also use Nix.

Kodlak15:
For what it's worth, I managed to find a way around the prior issue I was having with Go being a runtime dependency. The solution was rather simple. I created a shell application with Go as a runtime dependency, and that script just calls go-blueprint and passes any additional arguments to it. This way Go is available when the program is run, but does not need to be installed system-wide in order to run go-blueprint. It is basically equivalent to this shell script:
```
#!/usr/bin/env nix-shell
# nix-shell -p go

go-blueprint "$@"
```

The one thing I am still slightly unsure about is the best way of automating updating the vendor hash (in default.nix) when needed. It only needs to be updated if/when dependencies for the project change, which the hash doesn't seem to have changed since when I originally started working on this. The only way I know of to get the hash is to set `vendorHash` in default.nix to `lib.fakeHash`, run `nix build .#`, and retrieve the correct hash from the error output printed to screen. I can see if I can throw together a script that handles that automatically.

Ujstor:
Thanks for the update. We discussed internally whether we should accept this PR. The final decision is not yet made. Someone more familiar with Nix will check this PR, and then we can make a final decision.

Kodlak15:
Thanks to you as well for the update. Sounds great.

Ujstor:
Sorry for the big delay. Unfortunately, for now, we are not adding the flake config into the codebase. We are not Nix experts, and we cannot maintain it in the future if needed.

I hope you find another issue that can be solved and contribute that way :)

Kodlak15:
No worries. It was just something I was messing about with for my own purposes and figured I would share. I totally understand your rationale. Thanks for giving it some thought anyways!

PR #252, opened 2024-06-09
added middleware code to all frameworks

LrsK:

## Problem/Feature

This PR contains code to add middleware wiring for all frameworks, including example middleware based on the documentation for the respective framework.

## Description of Changes: 

- Added middleware file, in server package
- Added Middleware renderer to  Templater interface
- Middleware example and use in router is included by default in all frameworks

## Checklist

- [] I have updated the documentation (if applicable)

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -65,6 +65,7 @@ type Docker struct {
 type Templater interface {
 	Main() []byte
 	Server() []byte
+	Middleware() []byte
 	Routes() []byte
 	TestHandler() []byte
 	HtmxTemplRoutes() []byte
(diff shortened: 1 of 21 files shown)


briancbarrow:
Hey @LrsK, thanks for making a PR. We discussed it and we like the idea of adding basic middleware to all frameworks. One thing we'd like to see is consistency across the implementations of middleware. We are thinking you could add a simple logging middleware to everything. 

Echo already has their own logging middleware that logs info about the request, so maybe it can be something similar to that.

What do you think?

LrsK:
That sounds good. I can make a middleware called Logger or something, which would also follow the naming of Echo's logging middleware. I will just use the log/slog package in the standard library and log something simple.

This can replace the DummyMiddleware which was meant as an example that follows the docs from each framework.

briancbarrow:
> That sounds good. I can make a middleware called Logger or something, which would also follow the naming of Echo's logging middleware. I will just use the log/slog package in the standard library and log something simple.
> 
> This can replace the DummyMiddleware which was meant as an example that follows the docs from each framework.

That sounds great!

Ujstor:
@LrsK Sorry for the long, long delay. New features have been coming in, and there wasn't a clear picture of how to incorporate the middleware part into the codebase. Should we go with a simple implementation as it is, or something more complex.

Some frameworks, like Chi and Gin, already have logger middleware, which could cause conflicts. However, the more important part is ticket #253 —if this is incorporated correctly, we'll need a more complex, prod ready middleware implementation.

For now, I'm closing the PR, and I'll tag you once I open a middleware issue.

PR #254, opened 2024-06-10
bugfix: added special treatment of advanced flag in tip string

LrsK:

## Problem/Feature

The line that prints after the program exits. giving a quick command line version of the same options, had a bug in it that would show the --advanced true/false no matter what. This flag doesn't need a true or a false value to be passed in, only its inclusion is enough to go to advanced mode.

## Description of Changes: 

- Added check for all bool-flags, and changed the generated command accordingly. This is only applicable to advanced-flag, but will work for all bool-flags in the future.

## Checklist


Ujstor (approved):
LGTM

Issue #218, opened 2024-04-06
📋Update documentation if the PR introduces a new feature!📋


Ujstor:
### Tell us about your feature request

Updating official docs is mandatory if a new feature is introduced or existing ones are changed so that they are not documented in [docs.go-blueprint.dev](https://docs.go-blueprint.dev/)

Docs dir contains an MkDocs config that is easy to update with Markdown files. If there are any questions about the structure or needed content, this is the place to address these questions.

For local docs preview use python, install requirements.txt and run `make s`

### Disclaimer

- [X] I agree

PR #270, opened 2024-07-01
Add git options (commit, stage, and no git repository)

tomasohCHOM:

## Problem/Feature

Please include a description of the problem or feature this PR is addressing.

## Description of Changes: 

- Allow users to create a new project with the following Git options: init repo + commit the changes, init repo + stage the changes, proceed without creating a git repo
- Create `git` flag to allow for the same options as in the interactive UI
- Update the docs to include the new `git` flag

## Checklist

- [X] I have updated the documentation (check issue ticket #218)


tomasohCHOM:
I am pretty sure the workflow is failing because I added the extra flag `--git` and it will just not do anything since it expects a UI response otherwise. I pinpointed the file [`.github/workflows/generate-linter.yml`](https://github.com/Melkeydev/go-blueprint/blob/fef23e35e32bfe912e9fcfbdd53c58f3b04651cd/.github/workflows/generate-linter.yml#L8) where the action is being run. Would it be okay if I added `--git none` to fix the issue?

Ujstor:
@tomasohCHOM  You need to update the workflow because a new flag was introduced, and that is the reason why the workflow is failing

```yml
matrix:
        framework:
          [chi, gin, fiber, gorilla/mux, httprouter, standard-library, echo]
        driver:
          [mysql, postgres, sqlite, mongo, redis, none]
        git:
          [commit, stage, skip}
        advanced:
          [htmx, githubaction, websocket]
```


```yml
 - name: build templates
        run: script -q /dev/null -c "go run main.go create -n ${{ matrix.framework }} -f ${{ matrix.framework}} -d ${{matrix.driver}}  -g ${{matrix.git}}  --advanced true --feature ${{ matrix.advanced }}" /dev/null
```

Action will run the updated workflow on your PR.

tomasohCHOM:
@Ujstor for sure, I'll do that right now!

tomasohCHOM:
Damn it seems that adding those three more flag options will exceed the 256 matrix size limit it seems.
image

Ujstor:
Hmm, interesting. I didn't know what a final limit is. Maybe we can constrain the maximum parallel jobs? Not sure if this would work, but let's give it a try.

```yml
jobs:
  framework_matrix:
    strategy:
      max-parallel: 128
      matrix:
        framework:
         [chi, gin, fiber, gorilla/mux, httprouter, standard-library, echo]
        driver:
          [mysql, postgres, sqlite, mongo, redis, none]
        advanced:
          [htmx, githubaction, websocket]
        git:
          [commit, stage, skip]
```

tomasohCHOM:
No luck with that unfortunately. It says in the [GitHub documentation](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs):

> "A matrix will generate a maximum of 256 jobs per workflow run. This limit applies to both GitHub-hosted and self-hosted runners."

If we somehow can break it into multiple workflows, it might work. The flag that I added however does not contribute anything to the templates themselves, so if it is not 100% necessary, we could add a placeholder flag in `build templates` to at least make the workflow run, like:

```yaml
- name: build templates
        run: script -q /dev/null -c "go run main.go create -n ${{ matrix.framework }} -f ${{ matrix.framework}} -d ${{matrix.driver}}  -g skip  --advanced true --feature ${{ matrix.advanced }}" /dev/null
```

Ujstor:
> No luck with that unfortunately. It says in the [GitHub documentation](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs):
> 
> > "A matrix will generate a maximum of 256 jobs per workflow run. This limit applies to both GitHub-hosted and self-hosted runners."
> 
> If we somehow can break it into multiple workflows, it might work. The flag that I added however does not contribute anything to the templates themselves, so if it is not 100% necessary, we could add a placeholder flag in `build templates` to at least make the workflow run, like:
> 
> ```yaml
> - name: build templates
>         run: script -q /dev/null -c "go run main.go create -n ${{ matrix.framework }} -f ${{ matrix.framework}} -d ${{matrix.driver}}  -g skip  --advanced true --feature ${{ matrix.advanced }}" /dev/null
> ```

Breaking it into two pieces was also on my mind. For now, your proposal is okay, but better change flag to commit, that way, we can catch bugs and also remove max-parallel.

tomasohCHOM:
Got it, let me add that for now. Thank you!

Ujstor (changes requested):
Works as expected, nice PR. Only one small thing,  please add a new CLI screenshot to the root and docs public folder. After that, it is good to go.
blueprint_1

tomasohCHOM:
Thank you!! Will add it in just a second 👍

Ujstor:
@tomasohCHOM  Check also README.md file and add the --git commit flag in the appropriate places.

tomasohCHOM:
@Ujstor Ok I added the `--git commit` option for each of the snippets that needed it. I put the flag at the end of each snippet because it's the option that appears last in the UI. I also ensured that `docs` was updated with the new flag and it is through `docs/creating-project/project-init.md`

Ujstor:
I reran them manually, and now they are passing. More often than not, there is an issue with actions.

Ujstor (approved):
LGTM

Issue #282, opened 2024-07-23
[Bug] air.toml configuration will not work on windows


jexroid:
### What is the problem?

hi, first of all thanks for this awsome tool you made, I know you hate Windows but this air configuration will not work on Windows because of this MakeFile:10 `@go build -o main cmd/api/main.go`
it works fine on linux but on windows it should be `@go build -o tmp/main.exe cmd/api/main.go`
and air will not confuse which file to run cause If you don't specify .exe it will say `'main' is not recognized as an internal or external command,
operable program or batch file.` 

I am going to contribute on this project

### Operating System

windows 10

### Architecture Version (x86, x64, arm, etc)

x64

## detect the os and make the `MakeFile` based on that

I can contribute for fixing this error

on windows it should generate this:
```toml
#.air.toml
bin = ".\\main.exe"
cmd = "make build"
```
`MakeFile`
```shell
@go build -o main.exe cmd/api/main.go
```

instead of 
```toml
#.air.toml
bin = "./main"
cmd = "make build"
```
`MakeFile`
```shell
@go build -o main cmd/api/main.go
```

### Relevant log output

```shell
"Building..."

'E:\backend\gopi\main' is not recognized as an internal or external command,
operable program or batch file.
```

PR #283, opened 2024-07-23
[bug] live reload fixed

jexroid:

## Problem/Feature

air live reload will be working on both Unixbased and Windows machines

## Description of Changes: 

- Related to issue ticket #282 

1. `UnixBased` with bool type is added to `Project` type
2. CheckOs() function is added for checking the UNIX-based operating system if it's not windows leaves the `UnixBased` to false by default
```go
func (p *Project) CheckOs() {
	if runtime.GOOS != "windows" {
		p.UnixBased = true
	}
}
```
3. CheckOs() is executing in program.go:260
```go
// Define Operating system
p.CheckOs()
```
4. makefile.tmpl and air.toml.tmpl are modified

## Checklist

- [ ] I have updated the documentation (check issue ticket #218)

### DIFF
--- a/cmd/template/framework/files/makefile.tmpl
+++ b/cmd/template/framework/files/makefile.tmpl
@@ -7,7 +7,7 @@ build:
 	@echo "Building..."
 	{{if or ( .AdvancedOptions.htmx ) ( .AdvancedOptions.tailwind )}}@templ generate{{end}}
 	{{if .AdvancedOptions.tailwind}}@tailwindcss -i cmd/web/assets/css/input.css -o cmd/web/assets/css/output.css{{end}}
-	@go build -o main cmd/api/main.go
+	{{if .UnixBased }}@go build -o main cmd/api/main.go{{ else }}@go build -o main.exe cmd/api/main.go{{ end }}
 
 # Run the application
 run:
--- a/cmd/template/framework/files/makefile.tmpl
+++ b/cmd/template/framework/files/makefile.tmpl
@@ -51,9 +51,9 @@ clean:
 	@rm -f main
 
 # Live Reload
-watch:
 {{if .UnixBased}}
-    @if command -v air > /dev/null; then \
+watch:
+	@if command -v air > /dev/null; then \
             air; \
             echo "Watching...";\
         else \
@@ -66,7 +66,10 @@ watch:
                 echo "You chose not to install air. Exiting..."; \
                 exit 1; \
             fi; \
-        fi{{else}}	@air
+        fi
+{{else}}
+watch:
+	@air
 {{end}}
 
 .PHONY: all build run test clean watch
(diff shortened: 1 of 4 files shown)


Ujstor:
@jexroid 
Nice PR! This solves the issue of manually configuring Air to get it working in Windows. Is there a way we can get the make watch command to work in Win? Check if Air is installed and offer an installation option. Is there a PowerShell syntax that will do the same as in Unix-based systems? 

If not, we can use templating and just map make watch to air with a comment that Air can be installed with go install github.com/air-verse/air@latest

jexroid:
thanks @Ujstor I think im able to write a PowerShell for that which checks if air command return system error or not found. and return error statement that tell the user to install it your self or do the instructions for installing it using `GO get`

jexroid:
these failing tests are because of this:
```
{{if and (ne .DBDriver "none") (ne .DBDriver "sqlite")}}
# Integrations Tests for the application
itest:
	@echo "Running integration tests..."
	@go test ./internal/database -v
{{end}}
```

and the CI is testing for itest which will never exist in a project with no database or with sqlite
is there anything else needed @Ujstor

Ujstor:
I am a little bit busy...I will try to check the PR this weekend

jexroid:
alright, as we discussed the purpose of this PR is to generate a binary version of the go project based on the user OS which is configured in Makefile and run the file in .air.toml which solve the hot reloading in windows machines but there will be no Air auto installation on windows but its behave like before for Unix base machines

cmd/template/framework/files/makefile.tmpl
```diff
@@ -51,20 +51,25 @@ clean:
...
+                echo "Watching...";\
+            else \
+                echo "You chose not to install air. Exiting..."; \
+                exit 1; \
+            fi; \
+        fi
+{{else}}
+watch:
+	@air
+{{end}}
```
Ujstor, line 73: Indentation is wrong; this is why the tests are failing.

For local tests, run:

go build
go install

Then create the project. Test the features that you are implementing.

cmd/template/framework/files/makefile.tmpl
```diff
@@ -51,20 +51,25 @@ clean:
...
 
 .PHONY: all build run test clean
```
Ujstor, line 75: Nit: Update .PHONY with commands that are missing.

Ujstor (changes requested):
Check comments.

jexroid:
seems like the `Makefile:56: *** missing separator.  Stop.` occurs when you use spaces instead of tabs. I realized and fixed that while using Windows but couldn't test it on UNIX machines. the problems have been addressed and fixed. 
- **watch** command has been added to **.PHONY**
- Indentation of MakeFile is fixed

Ujstor (commented):
LGTM

PR #284, opened 2024-07-23
move handler_test.go to internal/server package

### LINKED ISSUE #280: [Fix] Move handler_test.go into internal/server dir

Common practice is to place test files in the same directory as the Go files they test. We can move handler_test.go into internal/server/routes_test.go and remove the tests directory from the root. Also, update the docs that refer to the old project tree.

This would be consistent with PR #271, which introduces the DB integration test.

### Disclaimer

- [X] I agree

eric-jacobson:

## Problem/Feature

Resolves #280 

## Description of Changes: 

- moved handler_test.go to internal/server package and renamed to routes_test.go
- updated docs referring to handler_test.go in the previous location

## Checklist

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -113,7 +113,6 @@ const (
 	internalServerPath   = "internal/server"
 	internalDatabasePath = "internal/database"
 	gitHubActionPath     = ".github/workflows"
-	testHandlerPath      = "tests"
 )
 
 // ExitCLI checks if the Project has been exited, and closes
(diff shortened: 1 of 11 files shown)


Ujstor (changes requested):
This looks good. Thanks for solving the open issue. Only make test needs to be updated.

```bash
test:
	@echo "Testing..."
-	@go test ./tests -v
+	@go test ./... -v
```

Ujstor (approved):
LGTM

Issue #285, opened 2024-07-25
[Feature Request] Add option to include slashes and dots in project name


wojcraft:
### Tell us about your feature request

It's not possible to enter project name as this: `github.com/wojcraft/project-name` - I've got used to this naming for learning and I see it also affects imports. Can you help me with that?

Here is screenshot of a input that needs to be modified in order to solve this request
![image](https://github.com/user-attachments/assets/29c29726-03c4-442f-a2aa-d9ffee1975c4)

### Disclaimer

- [X] I agree

Ujstor:
You are right, it is not possible to use '/' as input in the project name. If the go-blueprint create -n github.com/user/project-name command is executed, the CLI would create subdirectories after '/'. Thanks for addressing this....it needs to be fixed.

Ujstor:
This is an easy fix, the only thing that needs to be changed is the regex in the sanitizeInput() function (textinput.go):

```go
func sanitizeInput(input string) error {
	matched, err := regexp.Match("^[a-zA-Z0-9_\\-\\./]+$", []byte(input))
	if !matched {
		return fmt.Errorf("string violates the input regex pattern, err: %v", err)
	}
	return nil
}
```

wojcraft:
> This is an easy fix, the only thing that needs to be changed is the regex in the sanitizeInput() function (textinput.go):
> 
> ```go
> func sanitizeInput(input string) error {
> 	matched, err := regexp.Match("^[a-zA-Z0-9_\\-\\./]+$", []byte(input))
> 	if !matched {
> 		return fmt.Errorf("string violates the input regex pattern, err: %v", err)
> 	}
> 	return nil
> }
> ```

> /^[a-zA-Z0-9_\\-\\.\/]+$
I rather meant to make dots and slashes available at the same time. To allow urls like github.com/wojcraft/test

Ujstor:
I agree...I did just a quick test.

H0llyW00dzZ:
Wait, I think you don't have to explicitly use a regex because it's a byte.

wojcraft:
Dear @Ujstor,

So, are you going to implement this feature? I feel a little disoriented looking at this issue, so I'm just asking about a status of it.

If you want to know more details about what I mean in this quite small Feature Request, just let me know!

All the best!
Wojcraft

Ujstor:
The issue is open, and anyone can solve it and submit a PR :)

Patel-Raj:
[...]

PR #298, opened 2024-08-26
Accept module name as input to create command

Patel-Raj:

## Problem/Feature

go-blueprint create command was not working as expected if the user enters module name as an input rather than directory name.

Eg: go-blueprint create --name github.com/Patel-Raj/project1

This would create sub-directories github.com, Patel-Raj and project1 and then create the projects source code in project1 directory.

After this change, the Go project will be created at project1 as the root directory and the go.mod would have github.com/Patel-Raj/project1 as its module name.

Fixes: #285 

## Description of Changes: 

- Fetched directory name from the input string as the last token of the input when Split by /
- Relaxed the sanitization regex to allow / and . characters in the input.
- Used that the directory name to create the root directory.
- Added unit tests for the newly added util functions.

## Checklist

- [Y] I have self-reviewed the changes being requested
- [Y] I have updated the documentation (if applicable)

### DIFF
--- a/.github/workflows/generate-linter-advanced.yml
+++ b/.github/workflows/generate-linter-advanced.yml
@@ -30,19 +30,23 @@ jobs:
           git config --global user.name 'testname'
           git config --global user.email 'testemail@users.noreply.github.com'
 
+      - name: Set framework variable
+        id: set-proejct-directory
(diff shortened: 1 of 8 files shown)


Ujstor (approved):
Sorry for the long delay, I didn't have time for proper testing. The PR is amazing, and it LGTM!!

Issue #315, opened 2024-10-14
Cursor not visible when go-blueprint create fails


beeploop:
### What is the problem?

I can't see my cursor whenever an error occur during go-blueprint create. I can still type but the cursor is invisible.

### Operating System

Pop OS

### Architecture Version (x86, x64, arm, etc)

x64

### Steps to reproduce

Run go-blueprint create with unreliable or without internet connection.

Ujstor:
It is an old bug that slipped in when the Spinner was implemented. The Spinner is running in a separate goroutine, and if an error occurs, the terminal gets captured, and there is no visible input. We thought the bug was resolved, but obviously it isn't.

Feel free to let me know if anyone is interested in working on resolving this issue.

vinitparekh17:
@Ujstor i would like to pick this up

Ujstor:
Resolved in #318

PR #318, opened 2024-10-17
#315 fix: cursor error when go-blueprint create fails

vinitparekh17:

## Problem/Feature

This PR resolves issue #315 

## Description of Changes: 

In the `CreateMainFile` function located in `/cmd/program/program.go`, the error handling has been updated to ensure the terminal is properly released. Specifically:

- The function now returns an `err` so that the `spinner.ReleaseTerminal()` call can be executed ( create.go line 273 ), ensuring the terminal is released even in the event of an error.

This change helps prevent terminal lock-ups by ensuring proper cleanup

## Checklist

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -20,7 +20,6 @@ import (
 	"github.com/melkeydev/go-blueprint/cmd/template/docker"
 	"github.com/melkeydev/go-blueprint/cmd/template/framework"
 	"github.com/melkeydev/go-blueprint/cmd/utils"
-	"github.com/spf13/cobra"
 )
 
 // A Project contains the data for the project folder
@@ -244,7 +243,6 @@ func (p *Project) CreateMainFile() error {
 	// Check if user.email is set.
 	emailSet, err := utils.CheckGitConfig("user.email")
 	if err != nil {
-		cobra.CheckErr(err)
 		return err
 	}
 
@@ -276,7 +274,6 @@ func (p *Project) CreateMainFile() error {
 	err = utils.InitGoMod(p.ProjectName, projectPath)
 	if err != nil {
 		log.Printf("Could not initialize go.mod in new project %v\n", err)
-		cobra.CheckErr(err)
 		return err
 	}
 
@@ -285,7 +282,6 @@ func (p *Project) CreateMainFile() error {
 		err = utils.GoGetPackage(projectPath, p.FrameworkMap[p.ProjectType].packageName)
 		if err != nil {
 			log.Printf("Could not install go dependency for the chosen framework %v\n", err)
-			cobra.CheckErr(err)
 			return err
 		}
 	}
@@ -296,21 +292,18 @@ func (p *Project) CreateMainFile() error {
 		err = utils.GoGetPackage(projectPath, p.DBDriverMap[p.DBDriver].packageName)
 		if err != nil {
 			log.Printf("Could not install go dependency for chosen driver %v\n", err)
-			cobra.CheckErr(err)
 			return err
 		}
 
 		err = p.CreatePath(internalDatabasePath, projectPath)
 		if err != nil {
 			log.Printf("Error creating path: %s", internalDatabasePath)
-			cobra.CheckErr(err)
 			return err
 		}
 
 		err = p.CreateFileWithInjection(internalDatabasePath, projectPath, "database.go", "database")
 		if err != nil {
 			log.Printf("Error injecting database.go file: %v", err)
-			cobra.CheckErr(err)
 			return err
 		}
 
@@ -318,7 +311,6 @@ func (p *Project) CreateMainFile() error {
 			err = p.CreateFileWithInjection(internalDatabasePath, projectPath, "database_test.go", "integration-tests")
 			if err != nil {
 				log.Printf("Error injecting database_test.go file: %v", err)
-				cobra.CheckErr(err)
 				return err
 			}
 		}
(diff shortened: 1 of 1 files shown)


Ujstor:
Your PR didn't solve the issue. Try testing it, for example, by using a path where go.mod doesn't exist in program.go on purpose:

```go
	// Create go.mod
	err = utils.InitGoMod(p.ProjectName, gitHubActionPath)
	if err != nil {
		log.Printf("Could not initialize go.mod in new project %v\n", err)
		cobra.CheckErr(err)
		return err
	}
```
go build
go install

Create the project....blueprint errors out, and the terminal still doesn't display visible input.

vinitparekh17:
> Your PR didn't solve the issue. Try testing it, for example, by using a path where go.mod doesn't exist in program.go on purpose:
> 
> ```go
> 	// Create go.mod
> 	err = utils.InitGoMod(p.ProjectName, gitHubActionPath)
> 	if err != nil {
> 		log.Printf("Could not initialize go.mod in new project %v\n", err)
> 		cobra.CheckErr(err)
> 		return err
> 	}
> ```
> 
> go build go install
> 
> Create the project....blueprint errors out, and the terminal still doesn't display visible input.

I've tested again according to your way and confirmed it's working correctly.
Here I am attaching the screenshot as ref.

go-print

vinitparekh17:
> Your PR didn't solve the issue. Try testing it, for example, by using a path where go.mod doesn't exist in program.go on purpose:
> 
> ```go
> 	// Create go.mod
> 	err = utils.InitGoMod(p.ProjectName, gitHubActionPath)
> 	if err != nil {
> 		log.Printf("Could not initialize go.mod in new project %v\n", err)
> 		cobra.CheckErr(err)
> 		return err
> 	}
> ```
> 
> go build go install
> 
> Create the project....blueprint errors out, and the terminal still doesn't display visible input.

**Code snippet: `err = utils.InitGoMod(p.ProjectName, gitHubActionPath)`**

Thank you for the code snippet. I understand your concern about the potential bug scenario. However, this case would only occur if someone intentionally modifies the code to match the provided snippet, which would deviate from our intended implementation. Should we still add additional safeguards for such edge cases, or do you think the current implementation sufficiently handles the expected use cases?

vinitparekh17:
@Ujstor I would like to know your thoughts

Ujstor:
My point was that returning the "raw" error will not solve the issue. I stretched my example just to prove the point. I don't know what went wrong in #315, but it can happen, and users will experience the issue. If you completely remove the spinner from the codebase, the bug disappears. That's how I know the spinner itself causes this issue.

In the end, we have duplication in the codebase, there is no need to return an error and use cobra.CheckErr. The codebase is the result of work from many contributors, and things like this sneak in over time. Maybe it's time for a refactor ;)

vinitparekh17:
> My point was that returning the "raw" error will not solve the issue. I stretched my example just to prove the point. I don't know what went wrong in #315, but it can happen, and users will experience the issue. If you completely remove the spinner from the codebase, the bug disappears. That's how I know the spinner itself causes this issue.
> 
> In the end, we have duplication in the codebase, there is no need to return an error and use cobra.CheckErr. The codebase is the result of work from many contributors, and things like this sneak in over time. Maybe it's time for a refactor ;)

Indeed

Ujstor:
@vinitparekh17 Would you like to dig deeper into the problem and try to find a possible solution, or should I close the PR?

vinitparekh17:
> @vinitparekh17 Would you like to dig deeper into the problem and try to find a possible solution, or should I close the PR?

I would like to resolve this issue, might need some time and suggestions

vinitparekh17:
@Ujstor I've tested it with your snippet, and it's now working without breaking. I've just updated the error handling in **create.go** to use `cobra.CheckErr` instead of handling it in the **CreateMainFile** function.

vinitparekh17:
> @Ujstor I've tested it with your snippet, and it's now working without breaking. I've just updated the error handling in **create.go** to use `cobra.CheckErr` instead of handling it in the **CreateMainFile** function.


### simple explaination
`cobra.CheckErr` was printing an error message and exiting with code 1, which caused [this line](https://github.com/Melkeydev/go-blueprint/blob/9f230cdf0691501b51bc2b8b605325e6c37d3297/cmd/create.go#L273) to never be executed, leading to the bug we were facing.

Ujstor:
Unfortunately, this is not a valid solution. The cobra library is one of the core components, and it is used for a reason.

vinitparekh17:
> Unfortunately, this is not a valid solution. The cobra library is one of the core components, and it is used for a reason.

Indeed and I just made sure that we use `cobra.CheckErr` once at [create.go](https://github.com/Melkeydev/go-blueprint/blob/9f230cdf0691501b51bc2b8b605325e6c37d3297/cmd/create.go#L273) rather than multiple times in `program.go` so we can execute these lines before getting terminated 

```go
err = project.CreateMainFile()
		if err != nil {
			if releaseErr := spinner.ReleaseTerminal(); releaseErr != nil {
				log.Printf("Problem releasing terminal: %v", releaseErr)
			}
			log.Printf("Problem creating files for project. %v", err)
			cobra.CheckErr(textinput.CreateErrorInputModel(err).Err())
		}
		
```

I manually tested this with your snippet as well as by my own scenarios and it shows error as it was, before releasing cursor.

Ujstor:
Thanks for the feedback. I need to review this in detail, and I hope to manage it by the end of the week.

Ujstor (approved):
I checked and tested your implementation, and you are right, the error was not handled correctly in program.go. Very, very nice bug hunt!

LGTM!!

PR #343, opened 2024-11-25
Svelte init

arafays:

## Feature

Added support for init svelte as a frontend framework

## Description of Changes: 

- Added selecting a template option while creating vite project (not in the cli but in code so any future templates can be easily addded without updating the command `npm create vite@latest --template templateOption`)
- Renamed `CreateViteReactProject` to `CreateViteProject` so this can be used for any other vite related templates that are added in the future
- Added support for selecting Svelte in cli
- added svelte in tests
- fixed a small typo in the react documentation changed `whit` to `with`

## Checklist

this is my first go project/contribution ever just been watching content related to go never made any project as I am a front-end developer so let me know if the code can be improved or restuctured for any future contributions.


arafays:
- fixed another error in showing `index.html` in public folder from `react` project directory structure in docs.

cmd/program/program.go
```diff
@@ -866,35 +881,64 @@ func (p *Project) CreateViteReactProject(projectPath string) error {
...
+			}
+
+			if err := os.WriteFile(filepath.Join(srcDir, "App.svelte"), advanced.SvelteTailwindAppfile(), 0644); err != nil {
+				return fmt.Errorf("failed to write App.svelte template: %w", err)
 			}
 		}
 
 		// set to false to not re-do in next step
 		p.AdvancedOptions[string(flags.Tailwind)] = false
+	} else {
```
arafays, line 931: - moved the copying of template file in else block of tailwind as currently the app file was written once and then again it was overwritten if the tailwind one needed to be copied so now it is only written once.

Melkeydev:
Going to look into this on the weekend.
One thing though, we are potentially thinking of a creative way to consolidate these frontend frameworks, instead of making the --advanced flag options bloated

arafays:
@Melkeydev yup I watched the video and read few comments about people wanting svelte and I thought it would be easy replacing just the vite template. 

and being newbie at Golang i thought this would be a good starting point.

About the --advanced flag that is exactly what I thought when seeing the code I thought this was the beta features flag that maybe changed into the future.

arafays:
- added svelte in tests
and it is passing [here](https://github.com/Melkeydev/go-blueprint/actions/runs/12126094599/job/33807484465) 

I noticed there isn't a test for react + tailwind when I was checking if svelte + tailwind test will run or not as it will the same as react

arafays:
@Melkeydev fixed the`make run` command

Ujstor:
@arafays We need to rethink the current CLI flow and implementation. However, one certainty is that a frontend flag that is independent and isolated from advanced options is the most probable solution. Implementing this requires many changes, and the code needs to be rewritten. Hopefully, this will be resolved in Q1 2025. You can leave the PR open or convert it to a draft. Once we have a clear picture and the new flag is implemented, you can update your PR.

arafays:
@Ujstor converted to draft so that the project doesnt look abandoned :D I will keep a on lookout where I can help.

Blackwidow-sudo:
While adding svelte is nice to have, are there even enough standalone svelte users these days?
For me it makes more sense to add sveltekit with the static adapter for SPAs.

PR #357, opened 2024-12-16
docker-react-nginx-config

### LINKED ISSUE #339: [Feature Request] Golang to Serve Frontend Assets (like React) instead of Node.Js

I work on projects that require pretty high security standards, and I have yet to get Node.Js pushed through (maybe Deno in the future??). Has anyone tried deploying go-blueprint with React and without Node - letting Golang serve the frontend assets instead? 

This would be similar to the official React + ASP.NET or Vue + ASP.NET templates. Vite becomes a proxy server during development for DX reasons (HMR, tailwind, etc.). When it's time to deploy, Vite builds the frontend assets into a public folder, which Golang picks up and serves. All API endpoints will get served through 'https://example.com/api/<some endpoint>' (it would feel kinda like Next.Js).

It could simplify the deployment story because you wouldn't necessarily need docker-compose just to serve HTML/CSS/JS assets. One Golang server would do all the work, and latency could be reduced (depending on how the project is deployed).

Deploying without Node means you can't deploy any server-side JS packages. But go-blueprint is a Golang backend project - if you're deploying backend JS, you're probably doing it wrong.

If this seems interesting, I am happy to tinker and create a PR, but I don't want to expand this project's scope because of a specific deployment opinion or security need.

Any thoughts? 

Huge fan!! Go Blueprint is incredible!!

### Disclaimer

- [X] I agree

Ujstor:

## Problem

close #339 

## Description of Changes: 

As described in the linked issue, we should not use Node for serving static assets.

## Checklist

### DIFF
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -639,7 +639,6 @@ func (p *Project) CreateMainFile() error {
 			if err != nil {
 				return err
 			}
-
 		}
 	}
 
--- a/cmd/program/program.go
+++ b/cmd/program/program.go
@@ -610,6 +610,21 @@ func (p *Project) CreateMainFile() error {
 			return err
 		}
 
+		if p.AdvancedOptions[string(flags.React)] {
+			NginxConf, err := os.Create(filepath.Join(projectPath, "nginx.conf"))
+			if err != nil {
+				return err
+			}
+			defer NginxConf.Close()
+
+			// inject nginx.conf template
+			NginxConfTemplate := template.Must(template.New("nginx.conf").Parse(string(advanced.NginxConf())))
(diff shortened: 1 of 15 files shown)


valxntine:
Just out of curiosity - is there any plans to expand this functionality? I'd typically reach for Caddy rather than nginx, for example. There are also other options such as Envoy which I currently use at work. Are we against turning this into a flag and having a few options for the user? (in the future, not suggesting a re-write of this PR)

Ujstor:
@valxntine This is a great idea, and it could be done when we are redesigning the CLI. Frontend stuff will have a separate flag, and we can expand advanced flags when Docker and frontend are paired. Caddy and Traefik with automatic SSL are something that I would like to see implemented.

Issue #363, opened 2025-01-13
[Bug] Wrong type is returned for postgres test


silaselisha:
### What is the problem?

Postgres `mustStartPostgresContainer` expects to return a function of type `func(ctx context.Context, opts ...testcontainers.TerminateOption) error` but `func(context.Context) error` is returned.

### Operating System

Linux

### Architecture Version (x86, x64, arm, etc)

x64, x86

### Steps to reproduce

1. Create a Golang project configured to use Postgres Database
2. CD into the newly created project for file `database_test.go` in the database directory
3. On lines 42 and 47 the wrong type is returned 

### Relevant log output

```shell
cannot use dbContainer.Terminate (value of type func(ctx context.Context, opts ...testcontainers.TerminateOption) error) as func(context.Context) error value in return statementcompilerIncompatibleAssign
var dbContainer *postgres.PostgresContainer
```

PR #364, opened 2025-01-13
fix: return type for postgres tests template

silaselisha:

## Problem/Feature

Postgres mustStartPostgresContainer expects to return a function of type func(ctx context.Context, opts ...testcontainers.TerminateOption) error but func(context.Context) error is returned.

## Description of Changes: 

- Wrong type is returned for postgres test, hence changed the postgres template configuration to ensure that the right type is returned when generating postgres test file.

## Checklist

- [x] I have updated postgres.tmpl (check issue ticket #363)


Ujstor (approved):
LGTM

PR #380, opened 2025-02-27
Fixed minor interrupt signal bug

joncalhoun:

## Problem/Feature

The code prints a message stating that you can press `Ctrl+C` again to force the shutdown, but this doesn't actually work because the signal listener is stopped in a deferred function call which doesn't run until after the graceful shutdown is complete.

One way to verify this is to add a very slow handler (eg use a `time.Sleep`), visit that page, then press `Ctrl+C` multiple times to try to force the shutdown.

## Description of Changes: 

- Call `stop()` immediately after receiving an interrupt signal to unregister our behavior and revert to default behavior.
- Opted to leave the `defer stop()` call since it doesn't matter if we call it twice. It could be removed, but I tend to leave it to avoid future bugs.

## Checklist


Ujstor (approved):
LGTM

Ujstor (approved):
LGTM

PR #391, opened 2025-04-20
fix: dont install less than v4 tailwindcss

arafays:

## Problem/Feature

As the @tailwindcss/vite is installed and only supports v4 we are using --prefer-offline flag that can install probably install  tailwindcss v3 from cache with this we force to use cache only is tailwindcss version is greater than v4

## Description of Changes: 

Added constraint so it only installs greater than v4 in the cache
"tailwindcss@^4"

## Checklist


Ujstor (approved):
LGTM

PR #397, opened 2025-05-22
Added the option use just as a builder

eggplannt:

## Problem/Feature

Added the option to chose between just and make as a builder.

## Description of Changes: 

- Abstracted makefile into builder option
- Added Just as a builder option
- Created Just template

## Checklist


Melkeydev:
will review if passes the checks!

Ujstor:
https://github.com/actions/runner/issues/3868

eggplannt:
Ok I figured it out, since I added more options to create, it broke the linter

I not very familiar with GitHub actions and Linters, how would you go about fixing this

.github/workflows/generate-linter-core.yml
```diff
@@ -8,9 +8,11 @@ jobs:
   framework_matrix:
     strategy:
       matrix:
-        framework: [chi, gin, fiber, gorilla/mux, httprouter, standard-library, echo]
+        framework:
+          [chi, gin, fiber, gorilla/mux, httprouter, standard-library, echo]
         driver: [mysql, postgres, sqlite, mongo, redis, scylla, none]
         git: [commit, stage, skip]
+        builder: [just]
```
Ujstor, line 15: we need to add both of them: make and just

.github/workflows/generate-linter-core.yml
```diff
@@ -34,10 +36,10 @@ jobs:
         run: echo "PROJECT_DIRECTORY=${{ matrix.framework }}" | sed 's/\//-/g' >> $GITHUB_ENV
 
       - name: build templates
-        run: script -q /dev/null -c "go run main.go create -n ${{ env.PROJECT_DIRECTORY }} -f ${{ matrix.framework}} -d ${{ matrix.driver }} -g ${{ matrix.git}}"
+        run: script -q /dev/null -c "go run main.go create -n ${{ env.PROJECT_DIRECTORY }} -f ${{ matrix.framework}} -d ${{ matrix.driver }} -g ${{ matrix.git}} -b ${{ matrix.builder}}"
```
Ujstor, line 39: -b is reserved for backend in CLI refactor

cmd/create.go
```diff
@@ -52,11 +53,13 @@ func init() {
 	createCmd.Flags().BoolP("advanced", "a", false, "Get prompts for advanced features")
 	createCmd.Flags().Var(&advancedFeatures, "feature", fmt.Sprintf("Advanced feature to use. Allowed values: %s", strings.Join(flags.AllowedAdvancedFeatures, ", ")))
 	createCmd.Flags().VarP(&flagGit, "git", "g", fmt.Sprintf("Git to use. Allowed values: %s", strings.Join(flags.AllowedGitsOptions, ", ")))
+	createCmd.Flags().VarP(&flagBuilder, "builder", "b", fmt.Sprintf("Builder to use. Allowed values: %s", strings.Join(flags.AllowedBuilders, ", ")))
```
Ujstor, line 56: -b is reserved for backend. You need to find a new flag that fits the purpose

test/.air.toml
```diff
@@ -0,0 +1,46 @@
+root = "."
```
Ujstor, line 1: Why are the tests in the root of the project? Check my previous comment https://github.com/Melkeydev/go-blueprint/pull/397#issuecomment-2900660247

.github/workflows/generate-linter-advanced.yml
```diff
@@ -8,9 +8,11 @@ jobs:
   framework_matrix:
     strategy:
       matrix:
-        framework: [chi, gin, fiber, gorilla/mux, httprouter, standard-library, echo]
+        framework:
+          [chi, gin, fiber, gorilla/mux, httprouter, standard-library, echo]
         driver: [postgres]
         git: [commit]
+        builder: [just]
```
Ujstor, line 15: we need to add both of them: make and just

Ujstor (changes requested):
You need to add the docs for https://docs.go-blueprint.dev/

Once you add all the changes, I can test the implementation.

.github/workflows/generate-linter-core.yml
eggplannt: But this will cause the amount of permutations to be over 256

test/.air.toml
eggplannt: sorry i'll delete that

Ujstor: The tests are good, and if you can expand them, even better. But they belong to the templates, so they need to address the specific framework

.github/workflows/generate-linter-core.yml
Ujstor: What is the number of permutations if they are used only in generate-linter-advanced.yml?

test/.air.toml
eggplannt: No that's not even what they are, I accidentally created a project using go-blueprint called 'test' in the root directory

.github/workflows/generate-linter-core.yml
eggplannt: if I add them to advanced there re 84 permutations, so that works thanks.

Ujstor: Good to know. I think we need to use self-hosted runners in the k8s cluster once the limit is hit

cmd/create.go
eggplannt: k I changed it to -br

Ujstor:
There is one more update required when a new feature is implemented. There is a separate branch for the Blueprint UI — you can test it locally

https://github.com/Melkeydev/go-blueprint/tree/blueprint_ui/blueprint-ui

https://go-blueprint.dev/

Ujstor:
Does the user need to install Just on Linux, Windows, or macOS before using it? Or is it included in the base system on any of those platforms?

eggplannt:
The user has to install just

Ujstor:
> The user has to install just

This could be the issue. Until now, Blueprint worked out of the box, Makefile installed all the necessary dependencies, like Tailwind. Now we have a chicken-and-egg problem

eggplannt:
maybe it should be added as an advanced feature, cause docker isn't auto installed either

Ujstor:
> maybe it should be added as an advanced feature, cause docker isn't auto installed either

I would argue that every serious developer uses Docker, Podman, containerd, etc. Now that I know this will break the project from working without additional dependencies, I'm having second thoughts. I need to check with @Melkeydev how to proceed

eggplannt:
how would task be different, it also needs to be installed by the user no?

Ujstor:
> how would task be different, it also needs to be installed by the user no?

Yes, that’s the point. If we’re adding Makefile alternatives and forcing users to rely on external dependencies, then Taskfile makes more sense because it uses YAML and a different approach that can handle complex task orchestration. From what I understand, Just uses same syntax as Make. What’s the benefit? Make is preinstalled on most operating systems — at least on Unix-based ones

eggplannt:
Could also be a third option, so the options would be make, just, and task. Or do u think we should drop task all together.

Ujstor:
@eggplannt I need to have a consultation with Melkey, then we will make the final decision

Ujstor:
I decided not to accept the PR because we would be introducing another dependency. Make is the de facto standard

PR #402, opened 2025-05-29
added hot reload proxy for htmx

eggplannt:

## Problem/Feature

Added air proxy for htmx so that the page refreshes when air rebuilds

## Checklist

### DIFF
--- a/cmd/template/framework/files/air.toml.tmpl
+++ b/cmd/template/framework/files/air.toml.tmpl
@@ -44,3 +44,10 @@ tmp_dir = "tmp"
 [screen]
   clear_on_rebuild = false
   keep_scroll = true
+{{- if .AdvancedOptions.htmx}}
+
+[proxy]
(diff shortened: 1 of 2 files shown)


cmd/template/framework/files/air.toml.tmpl
```diff
@@ -44,3 +44,10 @@ tmp_dir = "tmp"
 [screen]
   clear_on_rebuild = false
   keep_scroll = true
+{{- if .AdvancedOptions.htmx}}
+
+[proxy]
+  proxy_port = 3000
+  app_port = 8080
```
Ujstor, line 51: The app port and proxy are hardcoded—can you retrieve them from the .env file instead?

eggplannt: the ports are hardcoded in .env too though, maybe we could make it a variable in the project object and pass it into both?

PR #420, opened 2025-07-21
Fix the wrong variable name in standard_library.go.tmpl

### LINKED ISSUE #419: Incorrect variable naming in standard_library.go.tmpl
Using 'NewServer' as a variable reference is incorrect. This should be 'newServer' (the local variable) instead of 'NewServer' (which refers to the function name).

https://github.com/search?q=repo%3AMelkeydev%2Fgo-blueprint%20NewServer&type=code#L25

shiponcs:
It fixes #419 

## Problem/Feature
Using 'NewServer' as a variable reference is incorrect. This should be 'newServer' (the local variable) instead of 'NewServer' (which refers to the function name).


shiponcs:
@Melkeydev

Ujstor (approved):
LGTM!

